PR #9, opened 2021-03-24
CI: add build targets for broker and operator

blockloop:


.github/workflows/go.yaml
```diff
@@ -29,6 +29,46 @@ jobs:
         only-new-issues: true
         args: --timeout=2m
 
+  build-controller:
+    name: test
+    runs-on: ubuntu-latest
```
ewolinetz, line 34: :astonished: no fedora/centos targets?

blockloop: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#github-hosted-runners
No options for that

ewolinetz (approved):

ewolinetz:
interesting, it looks like the checks didn't finish before it merged...
@blockloop any way we can gate that?

PR #10, opened 2021-03-25
make lint a non-requirement for specific make targets

blockloop:
Lint is not a hard requirement to build targets so it shouldn't be in the Makefile.
This allows us to have separate, asynchronous CI checks for lint and builds.


periklis (approved):
/lgtm

PR #15, opened 2021-03-30
Add API for the Loki stack

periklis:
This PR provides an implementation of the types and constants for the API discussed in the [enhancement proposal](https://github.com/openshift/enhancements/pull/712). It is **not** providing any glue code to reconcile the API to an actual Loki Stack deployment.


api/v1beta1/lokistack_types.go
```diff
@@ -17,30 +17,305 @@ limitations under the License.
...
+
+	// Limits defines the limits to be applied to log stream processing.
+	//
+	// +optional
+	Limits *LimitsSpec `json:"limits,omitempty"`
+
+	// Template defines the resource/limits/tolerations/nodeselectors per component
+	//
+	// +optional
+	Template *LokiTemplateSpec `json:"template,omitempty"`
```
ewolinetz, line 273: i feel like this gives way too much rope... we should either scale horizontally for users or recommend a different tshirt size instead of giving them access to the man behind the curtain.

api/v1beta1/lokistack_types.go
```diff
@@ -17,30 +17,305 @@ limitations under the License.
...
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// LokiStackSizeType declares the type for loki cluster scale outs.
+//
+// +kubebuilder:validation:Enum=OneXExtraSmallSize;OneXSmall;OneXMedium
+type LokiStackSizeType string
+
+const (
+	// OneXExtraSmallSize defines the size of a single Loki deployment
```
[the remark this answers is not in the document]
periklis: IMHO the missing `FIXME` line here is only to what each t-shirt size translates for the user experience. We talked about add here as well as later in official docs sentences like:
`Use this size for ingestion rate ~2TB/day and query load ~100rps`

Basically we want to set expectations for the user on selecting one size. The translation of each size in Replicas/Request/Limits should be reflect in the `LokiTemplateSpec`. This spec is the blueprint for generating our manifests. The user doesn't need to create such a blueprint in the first place as we do this for the user. However the user should be able to read the values from a place that is expected to include them namely the custom resource.

api/v1beta1/lokistack_types.go
periklis: AFAIU the whole point of providing T-Shirt size to the user is to pave the road for a great day1+2 experience and not restrict the user entirely. A full restriction especially for requests/limits/replicas, could have negative impact on day 1+2 experience, when the user's cluster is not able:
- To schedule Loki because of the high requirements we put on CPU/Mem to ensure the expectation we set (i.e. `Use this size for ingestion rate ~2TB/day and query load ~100rps`) are met.
- The cluster size does not provide nodes as big as we expect.

In addition the template should be amended by the operator to reflect the blueprint for each size if the user is not using it. This is informational and gives the user a handle from where to start to tune, e.g.:
- User wants to choose different limits when using `1x.medium` because his log volume hits the expectation on the ingestion path, but the queriers are not used as much as expected and this wastes memory on his cluster.

Updating the Template should be for the user a "do it at your own risk" operation w/o a need for a `managementState: Unmanaged` that blocks the reconciliation. Putting lower values than the selected size and expecting the same performance should be not restricted, but we should clearly state in the official docs that we don't recommend nor support such.

blockloop: I am very concerned with giving resource limits/requests to the user because it
inevitably leads to misuse and underprovisioning, which leads to bad behavior.
If we provide optimal sizing based on our research tasks then can't we encourage
users to scale out instead of up? What are the downsides to adding more
replicas? I lean toward providing uniform sizes where a size determines the
limits and requests. When a component is overloaded we can scale out to
relieve some of the load. If a stack is overprovisioned then the user used the
wrong size from the chart. Horizontal scaling also removes our concern for
overprovisioning a node, because it encourages smaller, more focused cattle
rather than large and bulky pets. I think the better option is to force an
unmanaged state to override our recommended CPU and memory requests and limits.

ewolinetz: I think I may have incorrectly expressed what I thought the t-shirt size should achieve...
Basically the thought is it would be specified by the user to tell the operator what is the component instance size we should be using. From there, the operator will just scale up as it needs based on the performance of the cluster and build in an upper bound to prevent us from taking over an entire k8s cluster.

The tricky part would still be how can a customer know which size is the most correct for them and I think if we provide a 'calculator' of some sort that can do something like:
* deploy something like fluentd as one off pods to each node
* evaluate the log rate (tail only) for a minute
* look at the log rate and figure out what capacity is needed
* recommend a t-shirt size based on the above

The nice thing is we can eventually build that into the operator itself so users would only need to tell us something like the `s3` end point to use. Also, if we have a consistent basis for calculating we would be able to look at our t-shirt sizes and know how accurate they are... for example if we recommend `small` to users and it always reaches max scale and still has performance issues, either `small` is too small or we should be recommending `medium` (the main point is it would be consistent and free from human error).

I understand your thought behind having the blueprint to give users an escape hatch, however it's been shown (brought up in a PM alignment meeting) that even if we put disclaimers in our docs about what can go wrong, if we state that something can be done, people will use it and break themselves... if its possible to do in the API we'll have to support it.

blockloop: > the operator will just scale ~up~ out as it needs
I agree that t-shirt sizes should be something we use to determine the per-instance, per-component CPU and memory requests and limits. I don't want to allow users to tune CPU and memory. In an ideal situation, the only user input we should accept is an S3 bucket and a PVC storage class. We should be able to determine the correct size based on their cluster logs. As Eric mentioned, we can start out with pre-release as a separate tool that deploys a log scraper to the cluster and then measures the amount of logging. We should be able to use this measurement to determine the size based on our recommended calculations. As we move closer to V1 we can merge this into the operator and have it size the cluster by itself. For now, we can separate the concerns by having a t-shirt size in the API and let the users choose. Ideally, they would run our tool first to discover the recommended size based on their current cluster.

periklis: This is an awesome discussion! It sorts out the misunderstanding of T-Shirt size format. I fully agree with what you propose, so let's drop `requests` and `limits` from the template. I propose however to keep the template for `Replicas`, `Tolerations` and `NodeSelector`. WDYT?

Let me summarize the gist of the above conversation about T-Shirt size so that I can adapt this into to enhancement proposal:

- Each t-shirt size declares a request and a limit per component, e.g. small means:
 - distributor: CPU -> 500ms (req), 1000ms (limit), Mem -> 1GB (req), 2GB (limit)
- Each declaration per t-shirt-size is a "standard" size per component that we want to scale only horizontally by adding new pods based on load.
- The operator will eventually read the load per components (e.g. in form of metrics) and make informed decisions for a t-shirt size, e.g.
  - Add more replicas for ingesters
  - Remove some replicas for queriers
  - Recommend in the status field or per alert that the user should consider selecting a bigger t-shirt size.

Did I capture the above correctly? I am removing the req/limits as a next step from this API and will adapt the enhancement proposal accordingly.

ewolinetz: `Tolerations` and `NodeSelector` sure. 

Do we need to give users the ability to tune `Replicas`? Doesn't the operator/cluster have an idea what a sane value is based on everything else it knows and it can make that decision?

Correct me if I'm wrong but I feel like this is something we could evaluate before we even get to the point of consuming performance metrics of the cluster. We know how scaled out the cluster is so we should know what we should have as a sane value... and it seems like if we set a certain replica value the cluster would naturally scale up to handle it in a weird circular manner...

periklis: @ewolinetz As per our last video call conversation, I captured the requirement to keep the `Replicas` field until we get performance measurement in the operator or any horizontal autoscaling resource if available and then deprecate/drop the field. To my understand we want users to scale the Loki components horizontally only.

How should the operator know if any Loki components should be scaled out w/o performance measurements? I don't quite get the point of a circular manner here.

blockloop (approved):
/lgtm

we can iterate as necessary after this initial merge. Thanks for the work, Peri.

PR #17, opened 2021-04-01
Add support to install the operator via OLM in OCP clusters

periklis:
This PR provides the manifests to install the loki-operator via OLM. Specifically it provides:
- A full compatible bundle with manifests and annotations to create bundle images.
- Support to scrape metrics of this operator on OpenShift via cluster-monitoring.
- Backwards compatibility support for the legacy `manifests` directory to enable CPaaS builds.
- Full operator-sdk integration for local development and test via OLM on a dev OCP cluster.

In the following each of the above is discussed/demostrated in detail:

### OLM Deployments via Operator Hub

Screenshot_2021-04-01 Installed Operators · Red Hat OpenShift Container Platform

### Metrics support via OpenShift cluster-monitoring

image
image

### Backwards compatible manifests directory

In general the old `manifests` directory on the top-level directory of our repository is a generated directories, except for the files (Probably all of them obsolete):
- `art.yaml`
- `loki-operator.package.yaml`
- `5.1.preview.1/image-references`

The `manifests` directory is populated from the `bundle` directory after `make bundle` is successfull. This means that we are always generating a full-workable bundle directory and copy the contents. In turn this approach ensures that we can easily drop the legacy `manifests` directory for the `bundle` directory.

### Full operator-sdk integration for local development/test via OLM

Assuming one has a developer OCP cluster, the present PR enables support to test the local build using the operator-sdk commands, i.e. (using a custom quay.io repository):

First of all we need a bundle image available on a public registry:
1. Generate the bundle: `make bundle IMG=quay.io/periklis/loki-operator:v0.0.1`
2. Create the bundle image: `make bundle-build BUNDLE_IMG=quay.io/periklis/loki-operator-bundle:v0.0.1`
3. Push the bundle image: `make oci-push IMG=quay.io/periklis/loki-operator-bundle:v0.0.1`

Second we need an operator image available on a public registry:
1. Build the operator image: `make oci-build IMG=quay.io/periklis/loki-operator:v0.0.1`
2. Push the operator image: `make oci-push IMG=quay.io/periklis/loki-operator:v0.0.1`

Next using the `operator-sdk` we can deploy and test the operator:
1. Create a namespace: `oc create ns openshift-logging`
2. Enable metrics: `oc label ns/openshift-logging openshift.io/cluster-monitoring=true --overwrite`
3. Switch namespace: `oc project openshift-logging`
4. Deploy the operator: `operator-sdk run bundle --install-mode OwnNamespace quay.io/periklis/loki-operator-bundle:v0.0.1`
```
INFO[0008] Successfully created registry pod: quay-io-periklis-loki-operator-bundle-0-0-1 
INFO[0008] Created CatalogSource: loki-operator-catalog 
INFO[0009] OperatorGroup "operator-sdk-og" created      
INFO[0009] Created Subscription: loki-operator-v0-0-1-sub 
INFO[0012] Approved InstallPlan install-t5ckj for the Subscription: loki-operator-v0-0-1-sub 
INFO[0012] Waiting for ClusterServiceVersion "openshift-logging/loki-operator.v0.0.1" to reach 'Succeeded' phase 
INFO[0012]   Waiting for ClusterServiceVersion "openshift-logging/loki-operator.v0.0.1" to appear 
INFO[0022]   Found ClusterServiceVersion "openshift-logging/loki-operator.v0.0.1" phase: Pending 
INFO[0025]   Found ClusterServiceVersion "openshift-logging/loki-operator.v0.0.1" phase: Installing 
INFO[0051]   Found ClusterServiceVersion "openshift-logging/loki-operator.v0.0.1" phase: Succeeded 
INFO[0051] OLM has successfully installed "loki-operator.v0.0.1" 
```
5. Delete the operator: `operator-sdk cleanup loki-operator`
```
INFO[0000] subscription "loki-operator-v0-0-1-sub" deleted 
INFO[0000] customresourcedefinition "lokistacks.loki.openshift.io" deleted 
INFO[0000] clusterserviceversion "loki-operator.v0.0.1" deleted 
INFO[0000] catalogsource "loki-operator-catalog" deleted 
INFO[0001] Operator "loki-operator" uninstalled 
```


Makefile
```diff
@@ -113,20 +120,30 @@ oci-build:
...
+	@cp bundle/manifests/loki-operator-controller-manager-metrics-service_v1_service.yaml manifests/${CLUSTER_LOGGING_VERSION}/loki-operator-controller-manager-metrics-service_v1_service.yaml
+	@cp bundle/manifests/loki-operator-manager-config_v1_configmap.yaml manifests/${CLUSTER_LOGGING_VERSION}/loki-operator-manager-config_v1_configmap.yaml
+	@cp bundle/manifests/loki-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml manifests/${CLUSTER_LOGGING_VERSION}/loki-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+	@cp bundle/manifests/loki-operator-metrics-reader_rbac.authorization.k8s.io_v1_clusterrole.yaml manifests/${CLUSTER_LOGGING_VERSION}/loki-operator-metrics-reader_rbac.authorization.k8s.io_v1_clusterrole.yaml
+	@cp bundle/manifests/loki-operator-prometheus_rbac.authorization.k8s.io_v1_role.yaml manifests/${CLUSTER_LOGGING_VERSION}/loki-operator-prometheus_rbac.authorization.k8s.io_v1_role.yaml
+	@cp bundle/manifests/loki-operator-prometheus_rbac.authorization.k8s.io_v1_rolebinding.yaml manifests/${CLUSTER_LOGGING_VERSION}/loki-operator-prometheus_rbac.authorization.k8s.io_v1_rolebinding.yaml
```
blockloop, line 140: Instead of hardcoding the `cp` commands for each individual file, we should make these into a make target that uses makefile idioms and caching. I have rewritten this rule like the following:

```
CLUSTER_LOGGING_VERSION ?= 5.1.preview.1

legacy-manifest-dir: manifests/$(CLUSTER_LOGGING_VERSION)/.keep
manifests/$(CLUSTER_LOGGING_VERSION)/.keep:
	@mkdir -p $(shell dirname $@) && touch $@

# legacy-manifest-files acts as an intermediary target by depending on all
# bundle/manifest files being copied to the manifests directory. The manifest
# yaml files are handled by the target immediately following this, where they
# simply cp source to target using a wildcard. Put simply, legacy-manifest-files
# maps to every yaml under bundle/manifest remapped to it's legacy path e.g.
#   'bundle/manifests/loki.openshift.io_lokistacks.yaml' -> 'manifests/5.1/loki.openshift.io_lokistacks.yaml'
# where manifests/5.1/loki.openshift.io_lokistacks.yaml is the target.
legacy-manifest-files: legacy-manifest-dir
legacy-manifest-files: legacy-csv
legacy-manifest-files: $(patsubst bundle/manifests/%.yaml, manifests/$(CLUSTER_LOGGING_VERSION)/%.yaml, $(wildcard bundle/manifests/*.yaml))
manifests/$(CLUSTER_LOGGING_VERSION)/%.yaml: bundle/manifests/%.yaml
	@cp -v bundle/manifests/$*.yaml $@

# The CSV does not work under the previous rule, because the
# CLUSTER_LOGGING_VERSION is part of the target filename so it requires it's
# own target
legacy-csv: manifests/$(CLUSTER_LOGGING_VERSION)/loki-operator.v$(CLUSTER_LOGGING_VERSION).clusterserviceversion.yaml
manifests/$(CLUSTER_LOGGING_VERSION)/loki-operator.v$(CLUSTER_LOGGING_VERSION).clusterserviceversion.yaml: bundle/manifests/loki-operator.clusterserviceversion.yaml
	@cp -v $^ $@

```

Notes: 

1. CLUSTER_LOGGING_VERSION should not have quotes because it invalidates the cache since Make does not understand quotes. It treats them as literal quotes and the targets will have embedded quotes in them, which will lead to misrepresentation of the target files. 
2. legacy-manifest-dir should be a _real_ target so that it does not get executed unless the target has changed. In GNU make a target _can technically_ be a directory, but in *nix based systems the timestamp of the directory is modified every time a file within the directory is updated. This invalidates the cache and breaks the chain. Instead, it's common to use a `.keep` file as the target whose timestamp is used for caching purposes. 
3. All yaml files under the bundle/manifests directory are created under a single, wildcard rule - except for the CSV. Make will handle the mapping of src -> dest. Because the CSV has the logging version in the filename, I created a separate target. The only caveat to this is that legacy-manifest-files will create the non-versioned CSV file. This _can_ be fix by using `$(filter-out ...)` to remove it from the src files, however, I don't think it's harmful to have `*.clusterserviceversion.yaml` and `*.<version>.clusterserviceversion.yaml`. If it is harmful then we can fix it.

bundle/manifests/loki-operator.clusterserviceversion.yaml
```diff
@@ -167,16 +225,22 @@ spec:
...
                 resources:
                   limits:
-                    cpu: 100m
-                    memory: 30Mi
+                    cpu: 200m
+                    memory: 128Mi
                   requests:
                     cpu: 100m
-                    memory: 20Mi
+                    memory: 64Mi
```
blockloop, line 232: I think we need to remove `resources` from the operator altogether. See https://github.com/openshift/elasticsearch-operator/pull/683/files for reference.

bundle/manifests/loki-operator.clusterserviceversion.yaml
```diff
@@ -218,8 +282,8 @@ spec:
   - name: Loki Operator
     url: https://loki-operator.domain
   maintainers:
-  - email: brejones@redhat.com
-    name: brett
+  - email: aos-logging@redhat.com
+    name: Red Hat, AOS Logging
```
blockloop, line 286: BOOO :smile:

config/manager/manager.yaml
```diff
@@ -48,9 +39,9 @@ spec:
...
         resources:
           limits:
-            cpu: 100m
-            memory: 30Mi
+            cpu: 200m
+            memory: 128Mi
           requests:
             cpu: 100m
-            memory: 20Mi
+            memory: 64Mi
```
blockloop, line 46: same comment as above on removing resources

blockloop (changes requested):
I left a few comments, mostly about the Makefile

Makefile
blockloop: The last line of the `bundle` target needs to - temporarily - be `$(MAKE) legacy-manifest-files` because of the order of operations. I will have a followup PR to make the `bundle` target an _actual_ target like this one which will fix this issue. For now, we can have a temporary fix by adding `$(MAKE) legacy-manifest-files` to the end of the `bundle` target.

periklis: Thank you so much for this explanation. I have always feared Makefile's text functions because I considered them harmful as per directories content change too often to spot misuse.

periklis:
@blockloop WDYT of adding shortcut commands in the makefile for the procedure described above in section `Full operator-sdk integration for local development/test via OLM`. I have been very careful using `$(MAKE)` in the past as recursive make considered harmful but I have not made any bad experiences so far. I.e.:

```make
olm-deploy-bundle:
	$(MAKE) bundle
	$(MAKE) bundle-build
	$(MAKE) oci-push IMG=$(BUNDLE_IMG)

olm-deploy-operator:
	$(MAKE) oci-build
	$(MAKE) oci-push

olm-deploy: olm-deploy-bundle olm-deploy-operator $(OPERATOR_SDK)
	oc create ns openshift-logging
	oc label ns/openshift-logging openshift.io/cluster-monitoring=true --overwrite
	oc project openshift-logging
	$(OPERATOR_SDK) run bundle --install-mode OwnNamespace $(BUNDLE_IMG)

olm-undeploy: $(OPERATOR_SDK)
	$(OPERATOR_SDK) cleanup loki-operator
	oc delete ns openshift-logging
```

blockloop:
I don't think you need to use `$(MAKE)`. What about this?

```
olm-deploy-bundle: bundle bundle-build oci-push
olm-deploy-operator: oci-build oci-push
```

periklis:
> I don't think you need to use `$(MAKE)`. What about this?
> 
> ```
> olm-deploy-bundle: bundle bundle-build oci-push
> olm-deploy-operator: oci-build oci-push
> ```
> 
> Other than that, LGTM

You are in general right, except for the following case where I need to replace an env variable:
```
olm-deploy-bundle: bundle bundle-build
	$(MAKE) oci-push IMG=$(BUNDLE_IMG)
```

periklis:
@blockloop PTAL again, I believe we are set for flight on this one.

Makefile
```diff
@@ -113,19 +127,74 @@ oci-build:
...
+# CLUSTER_LOGGING_VERSION is part of the target filename so it requires it's
+# own target
+legacy-csv: manifests/$(CLUSTER_LOGGING_VERSION)/loki-operator.v$(CLUSTER_LOGGING_VERSION).clusterserviceversion.yaml
+manifests/$(CLUSTER_LOGGING_VERSION)/loki-operator.v$(CLUSTER_LOGGING_VERSION).clusterserviceversion.yaml: bundle/manifests/loki-operator.clusterserviceversion.yaml
+	@cp -v $^ $@
+
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: manifests $(KUSTOMIZE) $(OPERATOR_SDK)
+bundle: manifests legacy-manifest-dir $(KUSTOMIZE) $(OPERATOR_SDK)
```
blockloop, line 157: nit: `legacy-manifest-dir` is a prereq of `legacy-manifest-files`. I don't think this is necessary.

Makefile
```diff
@@ -113,19 +127,74 @@ oci-build:
...
+# an OpenShift cluster selected via KUBECONFIG.
+.PHONY: olm-deploy
+ifeq ($(or $(findstring openshift-logging,$(IMG)),$(findstring openshift-logging,$(BUNDLE_IMG))),openshift-logging)
+olm-deploy:
+	$(error Set variable REGISTRY_ORG to use a custom container registry org account for local development)
+else
+olm-deploy: olm-deploy-bundle olm-deploy-operator $(OPERATOR_SDK)
+	oc create ns $(CLUSTER_LOGGING_NS)
+	oc label ns/$(CLUSTER_LOGGING_NS) openshift.io/cluster-monitoring=true --overwrite
+	oc project $(CLUSTER_LOGGING_NS)
```
blockloop, line 188: This assumes an OpenShift environment, which I don't think we want. You should prefer `kubectl` and fallback to `oc` the same way `OCI_RUNTIME` is set at the top of the Makefile. You'll also need to remove the `project` idiom that's specific to `oc` and use the namespace flag `(-n [namespace])` instead. I am running OLM in KiND.

Makefile
```diff
@@ -113,19 +127,74 @@ oci-build:
...
+.PHONY: olm-deploy-bundle
+olm-deploy-bundle: bundle bundle-build
+	$(MAKE) oci-push IMG=$(BUNDLE_IMG)
+
+# Build and push the operator image to a container registry.
+.PHONY: olm-deploy-operator
+olm-deploy-operator: oci-build oci-push
+
+# Deploy the operator bundle and the operator via OLM into
+# an OpenShift cluster selected via KUBECONFIG.
```
blockloop, line 179:
```suggestion
# a Kubernetes cluster selected via KUBECONFIG.
```

blockloop (approved):
/lgtm

PR #18, opened 2021-04-06
fix make bundle

blockloop:
This fixes a circular dependency created by #15 and adds a CI check for `make bundle`


blockloop:
@periklis there is an issue from the olm bundle that I can't figure out. The circular dep problem is solved here, but there will need to be some investigation into the `make bundle` error. I'll leave this to you for tomorrow. Since the `olm bundle` CI step is not required feel free to merge this and work in another branch or you can work on this branch.

ewolinetz (approved):

PR #19, opened 2021-04-07
Add support for full development cycle on kind and OLM

periklis:
This PR provides a dead simple local deployment workflow for the `loki-operator` based on kind and OLM. The user should be able to deploy the operator via OLM by simply:
```
$ make kind-olm-deploy
```
and remove the deployment via:
```
$ make kind-olm-cleanup
```

The intermediate make target to create a kind cluster and install OLM run only once. The user can delete the whole cluster simply by:
```
$ make kind-delete-cluster
```


Makefile
```diff
@@ -166,6 +172,45 @@ bundle: manifests $(KUSTOMIZE) $(OPERATOR_SDK)
...
+
+# Deploy the operator bundle and the operator via OLM
+# on local kind cluster.
+LOKI_OPERATOR_SECRET_NAME = loki-operator-metrics
+.PHONY: kind-olm-deploy
+ifneq ($(findstring $(KIND_CLUSTER_NAME),$(shell $(KIND) get clusters)),$(KIND_CLUSTER_NAME))
+kind-olm-deploy: kind-install-olm olm-prepare-deploy $(CFSSL) $(CFSSLJSON)
+else
+kind-olm-deploy: olm-prepare-deploy $(CFSSL) $(CFSSLJSON)
+endif
```
blockloop, line 200: This is going to execute `kind get clusters` every time **any** make target is executed.

blockloop (changes requested):
I think all of these changes need to be part of the bin/olm-install.sh script and not in the Makefile. This introduces way too much excess in the Makefile, making it difficult to follow or comprehend because of the layers of indirection. I don't imagine many people will be deploying this unless they are testing OLM-specific features locally. The Makefile should be used for builds and build-related things, not executing scripts.

periklis:
As per discussion on Google Meet, we conclude that these target are not going to be used regurarly or if at all, thus closing this PR. Users should use `make deploy/undeploy` and provide a cluster themselves.

PR #20, opened 2021-04-12
Ensure bundle dependencies are overwritable by env

periklis:
This change set ensures that operator bundle dependencies are overridable via enviroment variables (e.g. `LOKI_IMAGE`). This enables to pin the images to internal builds.

Ref: https://issues.redhat.com/browse/LOG-1228

### DIFF
--- a/internal/handlers/lokistack_create.go
+++ b/internal/handlers/lokistack_create.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"os"
 
 	"github.com/ViaQ/logerr/kverrors"
 	"github.com/ViaQ/logerr/log"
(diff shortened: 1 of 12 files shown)


internal/handlers/lokistack_create.go
```diff
@@ -29,10 +30,17 @@ func CreateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client) error
 		return kverrors.Wrap(err, "failed to lookup lokistack", "name", req.NamespacedName)
 	}
 
+	img := os.Getenv("LOKI_IMAGE")
+	if img == "" {
+		ll.Error(nil, "could not find LOKI_IMAGE env variable", "name", req.NamespacedName)
+		return nil
```
blockloop, line 36: Can we just default this so that we don't have to Setenv in our tests? It's easier if we _allow_ an override, but default to a known value that works such as the one you added in `os.Setenv`

blockloop (approved):
:+1:

We may pull this out of the create flow when the update flow is added in [LOG-1261](https://issues.redhat.com/browse/LOG-1261)

PR #24, opened 2021-04-15
Split default target into development and production overlays

periklis:
This PR provides a simplification for deploying the operator for dev purposes without requiring external dependencies:
- TLS cert/key pair for `kube-rbac-proxy` sidecar. I.e. the sidecar is removed from development deployments.
- Prometheus CRDs. I.e. the third-party resources like `ServiceMonitor` are removed from development deployments.

Note: The bundle generation (a.k.a. production manifests) are unaffected by this change.


blockloop (approved):
Nice work. Kustomize actually came in handy here :smile:

PR #27, opened 2021-04-16
Provide handler for LokiStack create and update events

periklis:
This PR provides a new combined handler for `LokiStack` create or update events. The general approach of the implementation relies on `k8s.io/apimachinery`'s `equality` pkg to test `LokiStack` owned objects (e.g. Statefulsets, ConfigMaps, Services).


internal/handlers/lokistack_create_or_update.go
```diff
@@ -51,30 +54,51 @@ func CreateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client) error
...
+			l.Error(err, "failed to set controller owner reference to resource", "resource", obj.GetName())
+			errCount++
+			continue
+		}
+
+		var existing client.Object
+		key := client.ObjectKeyFromObject(obj)
+		if err := k.Get(ctx, key, existing); err != nil {
+			if !apierrors.IsNotFound(err) {
+				errCount++
```
blockloop, line 76: I think you should log this error like you do elsewhere so that it's clear what is happening.

blockloop (commented):
Please add some unit tests for this new logic in `internal/handlers/lokistack_create_test.go`.

periklis:
@blockloop PTAL again.

blockloop:
I can't tell if you added update tests?

periklis:
> I can't tell if you added update tests?

Yeah, sorry renaming the `lokistack_create_test.go` to `lokistack_create_or_update_test.go` broke the diff experience. The following tests are new:
- `TestCreateOrUpdateLokiStack_WhenGetReturnsNoError_UpdateObjects`
- `TestCreateOrUpdateLokiStack_WhenSetControllerRefInvalid_ContinueWithOtherObjects`
- `TestCreateOrUpdateLokiStack_WhenCreateReturnsError_ContinueWithOtherObjects`
- `TestCreateOrUpdateLokiStack_WhenObjectsSemanticEqual_ContinueWithOtherObjects`
- `TestCreateOrUpdateLokiStack_WhenUpdateReturnsError_ContinueWithOtherObjects`

blockloop (commented):
This is nice and clean. Great work Peri. I'd like to see some tests on mutateFunc though to make sure we don't have any regressions in the future.

internal/manifests/mutate.go
```diff
@@ -0,0 +1,85 @@
...
+
+		case *appsv1.StatefulSet:
+			sts := existing.(*appsv1.StatefulSet)
+			wantSts := desired.(*appsv1.StatefulSet)
+			mutateStatefulSet(sts, wantSts)
+
+		default:
+			t := reflect.TypeOf(existing).String()
+			err := fmt.Errorf("missing mutate implementation for resource type %s", t) //nolint:goerr113
+			panic(err)
```
blockloop, line 49: You can't panic. We have to figure something else out here. Change the signature of MuteFuncFor to `controllerutil.MutateFn, error`

blockloop: https://golang.org/doc/effective_go#panic

This is not a compile-time issue that would always be caught. The tests capture the problem now, but panics _could_ happen in a production environment and this destroys any traceability we can gain from error handling such as logs, etc. Panics explicitly exit the application and bypass any upstream handling like closing clients, etc.

blockloop (approved):

PR #30, opened 2021-04-19
Add support for compactor

periklis:
This PR provides support for the Loki compactor component. It adds a new statefulset and two service (one for grpc, one for http metrics) to the manifests builds. In addition it ensures that only a single-replica compactor regardless user configuration.

### DIFF
--- a/internal/manifests/build.go
+++ b/internal/manifests/build.go
@@ -28,6 +28,7 @@ func BuildAll(opt Options) ([]client.Object, error) {
 	}
 	res = append(res, querierObjects...)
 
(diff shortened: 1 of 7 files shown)


blockloop (changes requested):
Can you add the compactor to [this table](https://github.com/ViaQ/loki-operator/blob/e16a365c66eb8e51eee6479c86cf2a0461721875/internal/manifests/service_test.go#L28) so that the label matches are tested and guaranteed?

blockloop (approved):

PR #31, opened 2021-04-19
Fix default CSV sample

periklis:
This PR provides a set of small fixes to make the default CSV sample from `config/samples` usable for local deployments:
- Fix typos in Loki config options: `MaxEntriesPerQuery` -> MaxEntriesLimitPerQuery`, `MaxLabelsPerSeries` -> `MaxLabelNamesPerSeries`
- Pin sample `size` to `OneXSmall`
- Pin storage class to `standard`
- Pin development overlay to use namespace `default` instead of `openshift-logging`. This enabled deployments on kind w/o the requirement to create `openshift-logging` manually.


blockloop (approved):
lgtm

PR #32, opened 2021-04-20
LOG 1250: Adding prometheus to project; Registering relevant stack metrics

Red-GV:
The following telemetry metrics have been added to the LokiStack

- The size of the cluster used
- The type of limits imposed on the cluster
- The usage of the compactor
- The number of streams used by the cluster

JIRA: https://issues.redhat.com/browse/LOG-1250


Makefile
```diff
@@ -160,7 +160,7 @@ olm-deploy-operator: oci-build oci-push
 # Deploy the operator bundle and the operator via OLM into
 # an Kubernetes cluster selected via KUBECONFIG.
 .PHONY: olm-deploy
-ifeq ($(or $(findstring openshift-logging,$(IMG)),$(findstring openshift-logging,$(BUNDLE_IMG))),openshift-logging)
+ifeq ($(oc $(findstring openshift-logging,$(IMG)),$(findstring openshift-logging,$(BUNDLE_IMG))),openshift-logging)
```
blockloop, line 163: I think this was a mistake

internal/handlers/lokistack_create.go
```diff
@@ -66,9 +67,48 @@ func CreateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client) error
 		l.Info("Resource created", "resource", obj.GetName())
 	}
 
+	RegisterLokiSpecMetrics(stack.Spec)
+
 	return nil
 }
 
+// Registers metrics based on the LokiStack spec
+func RegisterLokiSpecMetrics(spec lokiv1beta1.LokiStackSpec) {
```
periklis, line 76: FYI, `register` refers in prometheus metrics terminology to the one-time task to register the counters/gauges/histogram (e.g. as you do in `metrics.go`). I suggest to rename this function to `collectMetricsFor(spec lokiv1beta1.LokiStackSpec)`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+	lokiv1beta1 "github.com/ViaQ/loki-operator/api/v1beta1"
+)
+
+const (
+	CompactorEnabled  string = "compactor_enabled"
+	CompactorDisabled string = "compactor_disabled"
+	TenantLimits      string = "tenant_limits"
+	GlobalLimits      string = "global_limits"
+)
```
periklis, line 17: The above seem to me to be label values. I suggest to prefix them with `Label`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+var (
+	sizeTracker = prometheus.NewGaugeVec(
```
periklis, line 20: nit. `Tracker` is an unusual suffix for metrics. I suggest to replace `Tracker` with `Metric`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+			Name: "lokistack_cluster_sizes",
+			Help: "Number of clusters that are using a specific size configuration",
+		},
+		[]string{"size"},
+	)
+
+	globalStreamTracker = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "lokistack_max_global_streams",
+			Help: "Total of max global streams defined by the custom resource",
```
periklis, line 31: Consider [best practices](https://prometheus.io/docs/practices/naming/) on metric and label naming, e.g. if the description refers to a total then the metric name should probably look like `lokistack_max_global_streams_total`.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+		var isSize float64 = 0
+
+		if sizeType == size {
+			isSize = 1
+		}
+
+		sizeTracker.With(prometheus.Labels{
+			"size": SizeString(size),
+		}).Set(isSize)
+	}
```
periklis, line 94: The check for a size type is not idiomatic for golang. Consider the following:
```go
var value string
switch sizeType {
case lokiv1beta1.OneXExtraSmallSize:
  value = "1x.xsmall"
case lokiv1beta1.OneXSmall:
  value = "1x.small"
case lokiv1beta1.OneXMedium:
  value = "1x.medium"
}

sizeMetric.With(prometheus.Labels{"size": value}).Set(1)
```

In addition we do not need to register the zero values for a type that is not given. Each Spec can have only one Size type and thus we want to register this only.

periklis (changes requested):
Awesome work! :rocket: 

AFAICS most metrics are set as boolean-like values (e.g. 0 or 1). However the reconciliation loop is capable to handle multiple CRs. This fact would be hidden in the metrics, e.g. one CR with size small and one with size medium. The last one reconciled will set the counter/gauge. WDYT?

Makefile
Red-GV: This change? No, it was intentional. Is `or` a program I'm missing or something?

Red-GV:
@periklis Sure, I can modify this to accommodate for that. However, I just want to make sure that we are on the same page. 

>However the reconciliation loop is capable to handle multiple CRs.

So, during the loop, multiple specs will be reconciled simultaneously or at different times? The former is easy to fix up and the later is a little more annoying (unless the operator is restarting during some of these moments).

Makefile
periklis: @Red-GV No `or` is a makefile conditional.

blockloop:
The convention for registering prometheus metrics is to put them into `func init()` for each package they reside. This isn't harmful to any other program logic and it guarantees a single registration. You register the metrics with minimal, constant label values, and then you instantiate them later with values. When you export the metrics later you use `metric.WithLabelValues` see [here](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec.WithLabelValues)

internal/metrics/metrics.go
blockloop: The scope of this variable is wide so I recommend `maxGlobalStreamsMetric` for a name

internal/metrics/metrics.go
blockloop: See https://www.robustperception.io/on-the-naming-of-things

Red-GV:
@blockloop or @periklis I guess my confusion/concern comes more from this angle. (Might just be me not knowing)

- User starts off with 2 CRs (one with small and one with medium). [Medium: 1, Small: 1]
- User updates a CR to go to medium. [Medium: 2, Small: 1]

In that case, I believe I need to reset the counter unless we're keeping track of this over the lifetime of its usage. Or in the scenario, will the operator is going to restart, which would then make the counter accurate [Medium: 2, Small: 0].

Red-GV:
Should be ready for review again @blockloop, @periklis.

internal/handlers/lokistack_create_or_update.go
```diff
@@ -102,12 +107,65 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
+	isCompactorEnabled := false
+	if spec.Template != nil {
+		isCompactorEnabled = spec.Template.Compactor != nil
+	}
+
+	if err := metrics.IncCompactorMetrics(spec.Size, isCompactorEnabled); err != nil {
+		ll.Error(err, "Unable to increase the compactor enablement metric")
+	}
+
+	if spec.Limits != nil {
```
blockloop, line 145: Invert this logic and reduce nesting

internal/metrics/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	lokiv1beta1 "github.com/ViaQ/loki-operator/api/v1beta1"
+)
+
+const (
+	// SizeOneXExtraSmallLabel is a metric label for the size metric
+	SizeOneXExtraSmallLabel string = "1x.extra-small"
+	// SizeOneXSmallLabel is a metric label for the size metric
+	SizeOneXSmallLabel string = "1x.small"
+	// SizeOneXMediumLabel is a metric label for the size metric
+	SizeOneXMediumLabel string = "1x.medium"
```
blockloop, line 19: Why are these copied from the other package instead of referenced?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+var (
+	componentRestartMetric = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "lokistack_component_restart_total",
+			Help: "Number of times a component undergoes a restart",
+		},
+		[]string{"size", "component"},
+	)
```
periklis, line 46: Why do we need this metric? IIRC there is not acceptance criteria to count restarts

internal/metrics/metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	Querier LokiComponentMetricLabel = "querier"
+	// Ingester defines a metric label for usage of a component
+	Ingester LokiComponentMetricLabel = "ingester"
+	// Compactor defines a metric label for usage of a component
+	Compactor LokiComponentMetricLabel = "compactor"
+	// Distributor defines a metric label for usage of a component
+	Distributor LokiComponentMetricLabel = "distributor"
+	// QueryFrontend defines a metric label for usage of a component
+	QueryFrontend LokiComponentMetricLabel = "query_frontend"
+)
```
periklis, line 37: nit. Enumeration constant usually start with their type a Prefix, i.e. `LabelDefinedLimitsType`

internal/handlers/lokistack_create_or_update.go
```diff
@@ -102,12 +107,49 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
+
 		l.Info(fmt.Sprintf("Resource has been %s", op))
 	}
 
 	if errCount > 0 {
 		return kverrors.New("failed to configure lokistack resources", "name", req.NamespacedName)
 	}
 
+	if collectMetrics {
+		collectMetricsFor(&stack.Spec)
```
periklis, line 123: The injection point here to call `collectMetricsFor` looks flawed to me by the fact that it runs once per reconciliation loop for the same `LokiStackSpec`. IUC this means that we increment the counters/gauges every time we reconcile an update. For example, assuming the operator owns only one single `LokiStack` resource, every time we update this resource the counter for `lokistack_scale_out_size_total` will go up by one. However, this counter should stay to one as there is only one `LokiStack` resource for the given size.

periklis (changes requested):

internal/handlers/lokistack_create_or_update.go
```diff
@@ -102,12 +110,68 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
 		l.Info(fmt.Sprintf("Resource has been %s", op))
 	}
 
 	if errCount > 0 {
 		return kverrors.New("failed to configure lokistack resources", "name", req.NamespacedName)
 	}
 
+	// 1x.extra-small is used only for development, so the metrics will not
+	// be collected.
+	if string(opts.Stack.Size) != string(lokiv1beta1.SizeOneXExtraSmall) {
```
periklis, line 123: No need to cast to `string` here. Both sides are of the same type.

internal/handlers/lokistack_create_or_update.go
```diff
@@ -102,12 +110,68 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
+
+func collectMetrics(spec *lokiv1beta1.LokiStackSpec, stackName string) {
+	defaultSpec := manifests.DefaultLokiStackSpec(spec.Size)
+	sizes := []lokiv1beta1.LokiStackSizeType{lokiv1beta1.SizeOneXSmall, lokiv1beta1.SizeOneXMedium}
+
+	for _, size := range sizes {
+		var globalRate, tenantRate float64 = 0, 0
+		isUsingSize, isUsingCustomGlobalLimits, isUsingTenantLimits := metrics.ValueFalse, metrics.ValueFalse, metrics.ValueFalse
+
+		if string(spec.Size) == string(size) {
```
periklis, line 138: No need to cast to `string` here. Both sides are of the same type.

internal/handlers/lokistack_create_or_update.go
```diff
@@ -102,12 +110,68 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
 	return nil
 }
+
+func collectMetrics(spec *lokiv1beta1.LokiStackSpec, stackName string) {
+	defaultSpec := manifests.DefaultLokiStackSpec(spec.Size)
+	sizes := []lokiv1beta1.LokiStackSizeType{lokiv1beta1.SizeOneXSmall, lokiv1beta1.SizeOneXMedium}
+
+	for _, size := range sizes {
+		var globalRate, tenantRate float64 = 0, 0
+		isUsingSize, isUsingCustomGlobalLimits, isUsingTenantLimits := metrics.ValueFalse, metrics.ValueFalse, metrics.ValueFalse
```
periklis, line 136: nit. improve readability by using one style and multiple lines for init:
```go
var (
     globalRate float64 = 0.0
     tenantRate float64 = 0.0
     isUsingSize = metrics.ValueFalse
     isUsingCustomGlobalLimits = metrics.ValueFalse
     isUsingTenantLimits = metrics.ValueFalse
)
```

internal/manifests/build.go
```diff
@@ -35,11 +30,20 @@ func BuildAll(opt Options) ([]client.Object, error) {
...
+// DefaultLokiStackSpec returns the default configuration for a LokiStack of
+// the specified size
+func DefaultLokiStackSpec(size lokiv1beta1.LokiStackSizeType) *lokiv1beta1.LokiStackSpec {
+	defaults := internal.StackSizeTable[size]
+	return (&defaults).DeepCopy()
+}
+
+// ApplyBuildSettings manipulates the options to conform to
+// build specifications
+func ApplyBuildSettings(opt *Options) error {
```
periklis, line 42: Proposed rename to `ApplyDefaultSettings`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	lokiv1beta1 "github.com/ViaQ/loki-operator/api/v1beta1"
+)
+
+// UserDefinedLimitsTypeLabel defines a label that describes the type of limits
+// imposed on the cluster
+type UserDefinedLimitsTypeLabel string
```
periklis, line 12: Proposed rename `UserDefinedLimitsType`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+
+// MetricBooleanValue defines an int which describes a boolean state
+type MetricBooleanValue int
+
+const (
+	// ValueTrue describes the true state
+	ValueTrue MetricBooleanValue = 1
+	// ValueFalse describes the false state
+	ValueFalse MetricBooleanValue = 0
+)
```
periklis, line 29: No need for extra types here. Just use `0` and `1`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+
+var (
+	deploymentMetric = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "lokistack_deployment_total",
```
periklis, line 34: Suggest to rename to `lokistack_deployments` as per gauge used here.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+			Help: "Number of clusters that are deployed",
+		},
+		[]string{"size", "stack_id"},
+	)
+
+	userDefinedLimitsMetric = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "lokistack_user_defined_limits_total",
```
periklis, line 42: Same here `lokistack_user_defined_limits`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+			Help: "Number of clusters that are using user defined limits",
+		},
+		[]string{"size", "stack_id", "type"},
+	)
+
+	globalStreamLimitMetric = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "lokistack_global_stream_limit_total",
```
periklis, line 50: Same here `lokistack_global_stream_limit`

internal/handlers/lokistack_create_or_update.go
periklis: Still a string cast here

internal/metrics/metrics.go
periklis: We don't need constants here.

internal/handlers/lokistack_create_or_update.go
Red-GV: sorry, missed that one. should be good now.

periklis: Nope still casting to strings

Red-GV: weird. _should_ be fine now...

periklis (approved):
LGTM

PR #33, opened 2021-04-22
Add reconciliation for the s3 object storage secret

periklis:
This PR provides the switch from using `filesystem` as object storage to `s3`. It leverages the `LokiStack`'s API `ObjectStorageSpec` to pass the values to the `loki-config.yaml`. In addition:
- Reconciliation sets the CR status to `Degraded` if the secret is missing or invalid
- Local development overlays include a minimal `minio` installation.


internal/external/k8s/client.go
```diff
@@ -9,21 +9,35 @@ import (
...
 
-	Status() client.StatusWriter
-
 	RESTMapper() meta.RESTMapper
 	Scheme() *runtime.Scheme
+
+	Status() client.StatusWriter
+}
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . StatusWriter
```
blockloop, line 34: You don't need this line since you're using the `counterfeiter:generate` directive with the line 12 change.

internal/handlers/internal/secrets/secrets_test.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+	}
+	for _, tst := range table {
+		tst := tst
+		t.Run(tst.name, func(t *testing.T) {
+			t.Parallel()
+
+			_, err := secrets.Extract(tst.secret)
+			if err == nil && tst.wantErr {
+				t.Errorf("missing error: %s", err)
```
blockloop, line 71: `err` is always nil here so there's no need to print it. Also, you're not testing when `tst.wantErr` is false, but `err` is not nil

blockloop (changes requested):
Looks good for the most part. A few small changes. I'm not sure how I feel about having to run minio to deploy the operator. I understand that it's automated, but it adds another layer of complexity for a simple dev environment. For now we can assume it's low maintenance and won't become a burden we have to deal with. Later down the road, if we start having issues with minio becoming a neusance then we should remove it from the default development deploy and make it optional.

internal/handlers/internal/secrets/secrets_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	for _, tst := range table {
+		tst := tst
+		t.Run(tst.name, func(t *testing.T) {
+			t.Parallel()
+
+			_, err := secrets.Extract(tst.secret)
+			if !tst.wantErr {
+				require.NoError(t, err)
+			}
```
blockloop, line 73: Lol I think you flipped the logic so now you're not testing the inverse. I think you just need to add `if tst.wantErr { require.NotNil(t, err) }` after line 73

blockloop (approved):

PR #35, opened 2021-04-26
Add config map hash annotation for restarting components

periklis:
This PR provides a simple implementation to trigger restarts for Loki components on `loki-config.yaml` changes. In detail, the config map contents are hashed and latter hash is stored as an annotation into the pod template spec for each component deployment/annotation. The k8s schedulers re-schedules all pods on pod template annotation changes.


internal/manifests/var.go
```diff
@@ -15,6 +15,12 @@ const (
 	DefaultContainerImage = "docker.io/grafana/loki:2.2.1"
 )
 
+func commonAnnotations(opt Options) map[string]string {
```
blockloop, line 18: since `Options` is not needed, can you reduce the complexity of this and change the signature to `commonAnnotations(sha string)`?

internal/manifests/build.go
```diff
@@ -8,10 +8,11 @@ import (
 func BuildAll(opt Options) ([]client.Object, error) {
 	res := make([]client.Object, 0)
 
-	cm, err := LokiConfigMap(opt)
+	cm, sha1, err := LokiConfigMap(opt)
```
ewolinetz, line 11: out of curiosity -- why did you choose to add another return value vs creating a function just to return the sha for a specific configmap section?

internal/manifests/compactor.go
```diff
@@ -106,26 +107,27 @@ func NewCompactorStatefulSet(opt Options) *appsv1.StatefulSet {
...
-				MatchLabels: labels.Merge(compactorLabels, GossipLabels()),
+				MatchLabels: labels.Merge(l, GossipLabels()),
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:   fmt.Sprintf("loki-compactor-%s", opt.Name),
-					Labels: labels.Merge(compactorLabels, GossipLabels()),
+					Name:        fmt.Sprintf("loki-compactor-%s", opt.Name),
+					Labels:      labels.Merge(l, GossipLabels()),
+					Annotations: a,
```
ewolinetz, line 123: shouldn't this be higher up at the statefulset meta level, not the pod template?

internal/manifests/compactor.go
```diff
@@ -97,7 +97,8 @@ func NewCompactorStatefulSet(opt Options) *appsv1.StatefulSet {
 		},
 	}
 
-	compactorLabels := ComponentLabels("compactor", opt.Name)
+	l := ComponentLabels("compactor", opt.Name)
+	a := commonAnnotations(opt.ConfigSHA1)
```
ewolinetz, line 101: does this mean that for every component, we would have the same hash?
i thought we had discussed the idea that each component would have a hash based on it's specific configurations

blockloop:
LGTM. I'll let @ewolinetz handle the approval since he has open questions.

internal/manifests/build.go
blockloop: I think LokiConfigMap is the one that marshals the content into binary format and shoves it deep into the resulting Object. To run it outside of that context would require traversing the configmap (`cm.BinaryData["some unknown key"]`) and _knowing_ where to find that data. I think this fits perfectly within the scope of `LokiConfigMap`.

internal/manifests/compactor.go
periklis: You need this annotation to restart the pods only. Adding them like [kubectl](https://github.com/kubernetes/kubectl/blob/release-1.20/pkg/polymorphichelpers/objectrestarter.go#L32) on the pod template doesn't bother the k8s deployment controller just the scheduler.

internal/manifests/compactor.go
periklis: This would be a next step to split the configuration per component. Right now some configuration is shared among components (e.g. `storage_config` and `limits_config`). I suggest to flesh a per-component+shared hashing in a follow-up PR.

ewolinetz (approved):

PR #36, opened 2021-04-26
Add missing manifests tests

periklis:
This PR adds some tests for missing labels checks for `Spec.Selector` and `Spec.Template.Labels` for newly introduced components and for existing deployments


blockloop (approved):
:+1:

PR #37, opened 2021-04-26
Streamline API terms based on enhancement proposal

periklis:
This PR provides some streamlining fixes based on discussions [here](openshift/enhancements/pull/712):
- Drops separate `URL` field from `ObjectStorageSpec` in favor of using only a `Secret`.
- Renames fields using `perUser` to `perTenant` suffix. Tenant and user are used interchangeably in the Loki configuration. So we stick with tenant as the major term.
- Add field `ManagementState` defaulting to `Managed`.

### DIFF
--- a/api/v1beta1/lokistack_types.go
+++ b/api/v1beta1/lokistack_types.go
@@ -24,6 +24,21 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// ManagementStateType defines the type for CR management states.
+//
+// +kubebuilder:validation:Enum=Managed;Unmanaged
+type ManagementStateType string
+
+const (
+	// ManagementStateManaged when the LokiStack custom resource should be
+	// reconciled by the operator.
+	ManagementStateManaged ManagementStateType = "Managed"
+
+	// ManagementStateUnmanaged when the LokiStack custom resource should not be
+	// reconciled by the operator.
+	ManagementStateUnmanaged ManagementStateType = "Unmanaged"
+)
+
 // LokiStackSizeType declares the type for loki cluster scale outs.
 //
 // +kubebuilder:validation:Enum=SizeOneXExtraSmall;SizeOneXSmall;SizeOneXMedium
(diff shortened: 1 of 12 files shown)


api/v1beta1/lokistack_types.go
```diff
@@ -24,6 +24,21 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// ManagementStateType defines the type for CR management states.
+//
+// +kubebuilder:validation:Enum=Managed;Unmanaged
+type ManagementStateType string
```
blockloop, line 30: I want to see if we can get away from this idea altogether. I don't like the idea of an operator that's not an operator. If a user wants to go unmanaged then they should use the helm charts provided by Grafana Labs. Is there some past context that enforces this concept?

ewolinetz: My thought with why we would want to have the ability to set a cluster to `Unmanaged` is to give users the ability to stop LO from reconciling on their particular cluster in case manual steps need to be taken. I agree on your statement about leaving things unmanaged defeats the purpose of having an operator, this is really just meant to be an "escape hatch" to fix things and then set the cluster to managed again.

We would stray away from the idea of telling people to "scale down the operator to prevent this" because as the operator is currently configured it can respond to multiple Loki CRs, so scaling down the operator could be disruptive to other consumers of it/other clusters it manages.

ewolinetz: current slack discussion is to instead use the pattern:
```
OperationalState: <Paused|Running>
```

This follows the pattern for deployment rollouts and is more concise what it achieves than `Managed|Unmanaged`

periklis: As per current slack discussion, I feel that `Operational` is a bit weird to understand for a user. The current terminology for operators is an operator owns and controls a set of resources. That explains the requirement to put `Controller: true` on each `OwnerReference` of controlled resources.

In addition `Operational` has an ambiguous meaning reading the CR. Who's operations are meant by this? Operations of the LokiStack vs. Operations of Reconciling the CR? We mean latter, but that's not obvious for a non-expert user IMHO.

OTH `ManagementState` and `Managed|Unmanaged` is a well established term. If your set to unmanaged, the operator is not managing your CR, that's all. Simple and concise. Furthermore, since we use CRD defaulting to `Managed`. This field vanishes from users' eyes in general and we don't need to consider that users need to understand this field, when writing a simple `LokiStack` CR.

I argue to keep this as is and move forward.

blockloop: I'm fine with this naming scheme in general. I am more concerned that this may encourage users to use this as a permanent solution when it's intended to be temporary. We can figure out how to handle this separately. Is it possible to make the comment for this more clear? Perhaps explain what each state means and for what they are used?

blockloop (approved):

PR #38, opened 2021-04-26
LOG-1247: add support for tolerations

blockloop:


internal/manifests/tolerations_test.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"testing"
+
+	lokiv1beta1 "github.com/ViaQ/loki-operator/api/v1beta1"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestTolerationsAreSetForEachComponent(t *testing.T) {
```
ewolinetz, line 12: can we also add a test to make sure if for a particular CR component we omit it, it won't have a toleration?

ewolinetz (approved):

PR #39, opened 2021-04-27
LOG-1247: add node selectors support

blockloop:


Makefile
```diff
@@ -74,8 +74,8 @@ OCI_RUNTIME ?= $(shell which podman || which docker)
 ENVTEST_ASSETS_DIR=$(CURDIR)/testbin
 test: generate go-generate lint manifests
 test: $(GO_FILES)
-	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.7.0/hack/setup-envtest.sh
+	@mkdir -p ${ENVTEST_ASSETS_DIR}
+	@test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.7.0/hack/setup-envtest.sh
```
blockloop, line 78: These are verbose when running `make test` and it's unnecessary output.

periklis: Do we need the `envtest` at all? We don't really use it right now. Why not dropping this?

periklis (approved):
LGTM

PR #41, opened 2021-04-30
Use pointers for optional API fields

periklis:
Based on the established [k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) optional fields w/o a built-in `nil` type should use pointers. This enables to distinguish the empty from the not-set case. Hence this PR is providing back the pointer types in the `LokiStack` API. 

In addition this PR provides a crucial fix uncovered by using pointers namely the template was per-default unset when building the manifests. The change set makes `manifests.Build` to use a copy of the `StackSizeTable[size]` and apply the user options on it. This ensures that manifests and config generation have access to a fully declared `LokiStackSpec` incl. `Limits` and `Template` fields. In turn this enables some goodies:
- Tests need to declare only required fields.
- No need to check for pointers when reading values in `loki-config.yaml`

In summary the user can declare a minimal `LokiStack` CR like this:
```yaml
apiVersion: loki.openshift.io/v1beta1
kind: LokiStack
metadata:
  name: lokistack-minimal
spec:
  size: 1x.extra-small
  replicationFactor: 1
  storageClassName: standard
  storage:
    secret:
      name: test
```


internal/manifests/build.go
```diff
@@ -1,37 +1,60 @@
...
+	spec := (&defs).DeepCopy()
+	if err := mergo.Merge(spec, opt.Stack, mergo.WithOverride); err != nil {
+		return Options{}, kverrors.Wrap(err, "failed merging stack user options", "name", opt.Name)
+	}
+
+	strictOverrides := lokiv1beta1.LokiStackSpec{
+		Template: &lokiv1beta1.LokiTemplateSpec{
+			Compactor: &lokiv1beta1.LokiComponentSpec{
+				// Compactor is a singelton application.
+				// Only on replica allowed!!!
```
blockloop, line 49: typo: only *one* replica allowed :)

blockloop (approved):

PR #44, opened 2021-05-04
Allowing controller to access secrets

Red-GV:
Adding the secrets to the RBAC role for the controller.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -23,6 +23,14 @@ rules:
   - update
   - watch
 - apiGroups:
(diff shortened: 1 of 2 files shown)


periklis:
> Adding the secrets to the RBAC role for the controller.

As per not reconciling owned secrets, I recommend to add a more fine-grained rule for secrets allowing only `get,watch,list`.

Red-GV:
Should be good to go now, @periklis. Thanks!

periklis (approved):
LGTM

PR #45, opened 2021-05-05
Add support for runtime config

periklis:
This PR adds reconciliation support for a [runtime-config](https://grafana.com/docs/loki/latest/configuration/#runtime-configuration-file) for all Loki components. The `runtime-config` represents a mean to provide tenant based ingestion/query limits. In detail the implementation adds supports:
- Parse and render the following `LokiStackSpec` section into a separate `runtime-config.yaml`.
- Conditional rendering of limits per tenant. Renders only non-empty values.
- Persist the runtime config file into the `LokiConfigMap` as separate binary blob. 
- All components load this file via the argument `-runtime-config.file`.
- ~A `SHA1` hash is calculated based on the runtime config contents and passed down to components. Any changes of this SHA1 triggers a restart of all components.~

LokiStackSpec:
```yaml
apiVersion: loki.openshift.io/v1beta1
kind: LokiStack
spec:
  ...
  limits:
    tenants:
      tenantA:
        ingestion:
          ingestionRate: 100
          ingestionBurstSize: 200
          maxLabelNameLength: 256
          maxLabelValueLength: 1024
          maxLabelNamesPerSeries: 100
          maxStreamsPerTenant: 1000
          maxGlobalStreamsPerTenant: 2000
        queries:
          maxEntriesLimitPerQuery: 100
          maxChunksPerQuery: 1000
          maxQuerySeries: 2000
      tenantB:
        ingestion:
          ingestionRate: 100
          ingestionBurstSize: 200
          maxLabelNameLength: 256
          maxLabelValueLength: 1024
          maxLabelNamesPerSeries: 100
          maxStreamsPerTenant: 1000
          maxGlobalStreamsPerTenant: 2000
          maxLineSize: 512
```

Runtime config file:
```yaml
overrides:
  tenantA:
    ingestion_rate_mb: 100
    ingestion_burst_size_mb: 200
    max_label_name_length: 256
    max_label_value_length: 1024
    max_label_names_per_series: 100
    max_streams_per_user: 1000
    max_global_streams_per_user: 2000
    max_entries_limit_per_query: 100
    max_chunks_per_query: 1000
    max_query_series: 2000
  tenantB:
    ingestion_rate_mb: 100
    ingestion_burst_size_mb: 200
    max_label_name_length: 256
    max_label_value_length: 1024
    max_label_names_per_series: 100
    max_streams_per_user: 1000
    max_line_size: 512
    max_global_streams_per_user: 2000

```

### DIFF
--- a/internal/manifests/build.go
+++ b/internal/manifests/build.go
@@ -18,11 +18,11 @@ func BuildAll(opt Options) ([]client.Object, error) {
 		return nil, err
 	}
 
-	cm, sha1, err := LokiConfigMap(opt)
+	cm, sha1C, err := LokiConfigMap(opt)
 	if err != nil {
 		return nil, err
 	}
-	opt.ConfigSHA1 = sha1
+	opt.ConfigSHA1 = sha1C
 
 	res = append(res, cm)
 	res = append(res, BuildDistributor(opt)...)
--- a/internal/manifests/compactor.go
+++ b/internal/manifests/compactor.go
@@ -48,6 +48,7 @@ func NewCompactorStatefulSet(opt Options) *appsv1.StatefulSet {
 				Args: []string{
 					"-target=compactor",
 					fmt.Sprintf("-config.file=%s", path.Join(config.LokiConfigMountDir, config.LokiConfigFileName)),
+					fmt.Sprintf("-runtime-config.file=%s", path.Join(config.LokiConfigMountDir, config.LokiRuntimeConfigFileName)),
 				},
 				ReadinessProbe: &corev1.Probe{
 					Handler: corev1.Handler{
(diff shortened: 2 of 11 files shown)


blockloop:
> A SHA1 hash is calculated based on the runtime config contents and passed down to components. Any changes of this SHA1 triggers a restart of all components.

ConfigMap volumes are automatically persisted to the volume when they are updated (unless we use SubPath). There's no need for this excess complexity. 

> When a ConfigMap already being consumed in a volume is updated, projected keys are eventually updated as well. Kubelet is checking whether the mounted ConfigMap is fresh on every periodic sync. However, it is using its local ttl-based cache for getting the current value of the ConfigMap. As a result, the total delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the pod can be as long as kubelet sync period (1 minute by default) + ttl of ConfigMaps cache (1 minute by default) in kubelet. You can trigger an immediate refresh by updating one of the pod's annotations.

Source: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

blockloop (approved):

PR #47, opened 2021-05-07
[SPIKE] Add support for smart restarts

periklis:
### Problem statement
The `LokiStack` CR allows only an opinionated list of knobs to configure the operator-managed Loki instances. In total there are three distinct configuration groups namely `ingester.lifecycler.ring.replication_factor`, `limits_config`, `storage_config`. The first group affects only `ingesters`, the second all except `compactor` and the third affects all components. Hence the problem at hand is to define a mechanism in the loki operator to restart only affected components. In turn this enables almost zero-disruption for non-affected components.

### Proposed implementation
This spike considers two configuration sources to keep the operator stateless as is. The one source is the contents persistent in the Loki ConfigMap and represents the current state of the union. The other source is the generated **but** not persisted ConfigMap from the `LokiStackSpec` and represents the desired state. The high level overview of the implementation is to compare both configuration states and provide a mapping of hashes for all component, where affected components have an updated hash. These hashes are applied to pod template annotations which the k8s scheduler considers for restarting the pods.

In summary the implementation provides the following steps:
1. Unmarshal both config contents in partial `Loki` config objects.
2. Compare them with [google-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp) using a custom reporter to register the roots of changed paths only.
3. Map the roots to a static list of affected components.
4. Return only a hash of the new config for affected components.
5. Return the old hash for non-affected components.

### Alternative approaches
The following alternatives have been considered but withdrawn:
1. Import grafana/loki as go-module to omit implementing custom loki config objects. **Withdrawn:** Loki does not follow go module conventions and in turn is not a meant to be vendored.
2. Looking for copy/paste-able sources in grafana/loki to unmarshal YAML into them. **Withdrawn:** As per Loki config objects are scattered across the upstream source code. Identification of all of them is error-prone and provides diminishing returns for supporting only an opinionated subset.

### Open questions
Although the exploration here aims to enable partial restarts of affected components, further questions remain if the proposed work and current implementation is the right mix for the operator:
1. How to maintain a static mapping of config keys to components across loki versions? Is this only a matter of another level of indirection?
2. What is the impact of deprecated & removed keys from the loki configuration? How does this impact the upgrade path in the operator?
3. Are there any hidden gems in the Loki configuration which demand to restart the whole cluster but not identifiable from the present configuration grouping?

cc @cyriltovena @slim-bean @owen-d

### DIFF
--- a/internal/handlers/lokistack_create_or_update.go
+++ b/internal/handlers/lokistack_create_or_update.go
@@ -60,6 +60,14 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
 		)
 	}
 
+	var cm corev1.ConfigMap
+	key = client.ObjectKey{Name: manifests.LokiConfigMapName(stack.Name), Namespace: stack.Namespace}
+	if err = k.Get(ctx, key, &cm); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return kverrors.Wrap(err, "failed to lookup lokistack config map", "name", key)
+		}
+	}
+
 	// Here we will translate the lokiv1beta1.LokiStack options into manifest options
 	opts := manifests.Options{
 		Name:          req.Name,
@@ -67,6 +75,10 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
 		Image:         img,
 		Stack:         stack.Spec,
 		ObjectStorage: *storage,
+		Config: manifests.Config{
+			Config:        cm.BinaryData["config.yaml"],
+			RuntimeConfig: cm.BinaryData["runtime-config.yml"],
+		},
 	}
 
 	ll.Info("begin building manifests")
--- a/internal/manifests/config.go
+++ b/internal/manifests/config.go
@@ -1,7 +1,6 @@
 package manifests
 
 import (
-	"crypto/sha1"
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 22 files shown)


internal/handlers/lokistack_create_or_update.go
```diff
@@ -60,13 +60,25 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
 	opts := manifests.Options{
 		Name:          req.Name,
 		Namespace:     req.Namespace,
 		Image:         img,
 		Stack:         stack.Spec,
 		ObjectStorage: *storage,
+		Config: manifests.Config{
+			Config:        cm.BinaryData["config.yaml"],
+			RuntimeConfig: cm.BinaryData["runtime-config.yml"],
+		},
```
ewolinetz, line 81: out of curiosity, why did you opt to add the contents of the CM into the config here vs just pulling it as part of `LokiConfigMap` call? it seems like we don't ever refer back to this data except for when we do our diff comparison...

internal/manifests/config.go
```diff
@@ -11,34 +10,32 @@ import (
...
 	cfg := ConfigOptions(opt)
 	c, rc, err := config.Build(cfg)
 	if err != nil {
-		return nil, "", err
+		return nil, nil, err
 	}
 
-	s := sha1.New()
-	_, err = s.Write(c)
+	res, err := config.Compare(opt.Config.Config, c)
```
ewolinetz, line 20: do we actually need to know what the current config is? wouldn't just knowing the resulting hashes for what we generate be enough?

periklis:
As per discussion closing this PR because we have present consensus on:
1. Selective/Smart restarts based on configuration changes is an important expert-level feature which we would like to see in the operator. Thus we don't drop the idea per se.
2. Selecting  Loki components to restart based on configuration changes is not an easy task to maintain because we need currently to maintain static mappings of configuration sections to components. Maybe this is a good point to seek to upstream for exposing configuration and types as a separate go module?! cc @cyriltovena)
3. We drop the present PR because 2 out of 3 configuration group represent full cluster restarts. Thus maintaining selecting restart only for `replication_factor` is not worth the effort and maintenance.

PR #50, opened 2021-05-11
Add support for a per component pod status map 

periklis:
### Context
The `LokiStack` custom resource represents an artificial resource that is not per se scheduled as a single container on a kubernetes cluster. However, a `LokiStack` owns a set of deployments (`distributor`, `query-frontend`) and a set of statefulsets (`compactor`, `ingester`, `query-frontend`) which in turn own the running pods of a Loki cluster. Hence providing an overview of pod status per component in the `LokiStackStatus` results in a one-place-to-stop and check the cluster installation. In addition providing the necessary operator-sdk markers enables an enhanced visualization using live donut-diagrams in the operand detail page (See [reference](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#1-podStatuses).

### Implementation
The proposed implementation provides a map of pod status per component in the `LokiStackStatus` struct. On each reconciliation loop this status field is re-created and persisted anew. In addition the controller is enabled to react on update events of owned sub-resources, i.e. for Deployments and StatefulSets only.

**Note:** Rendering of multiple donut-diagrams for multiple `podStatuses` descriptors is broken in OCP. See https://github.com/openshift/console/pull/8893

### Screenshots
image
image


controllers/internal/status/components.go
```diff
@@ -0,0 +1,96 @@
...
+
+	var (
+		compactorName     = manifests.CompactorName(s.Name)
+		distributorName   = manifests.DistributorName(s.Name)
+		ingesterName      = manifests.IngesterName(s.Name)
+		querierName       = manifests.QuerierName(s.Name)
+		queryFrontendName = manifests.QueryFrontendName(s.Name)
+		psm               = lokiv1beta1.PodStatusMap{}
+	)
+	type componentMapping struct {
```
blockloop, line 36: If there's no reason to define the struct within the func scope I think you should move it out.

controllers/internal/status/components.go
```diff
@@ -0,0 +1,96 @@
...
+		label string
+		psm   lokiv1beta1.PodStatusMap
+	}
+	components := map[string]componentMapping{
+		compactorName: {
+			label: manifests.LabelCompactorComponent,
+			psm:   psm.DeepCopy(),
```
blockloop, line 43: What's the point of using DeepCopy? I think you should just use a `lokiv1beta1.PodStatusMap{}` since `psm` is an empty struct. It looks like you're only using `psm` for this reason.

controllers/internal/status/components.go
```diff
@@ -0,0 +1,96 @@
...
+	status.Components = lokiv1beta1.LokiStackComponentStatus{
+		Compactor:     components[compactorName].psm,
+		Distributor:   components[distributorName].psm,
+		Ingester:      components[ingesterName].psm,
+		Querier:       components[querierName].psm,
+		QueryFrontend: components[queryFrontendName].psm,
+	}
+	s.Status = *status
+	return k.Status().Update(ctx, &s, &client.UpdateOptions{})
+}
```
blockloop, line 80: This whole function is far too complex to avoid a little repetition. It's confusing what is actually happening when trying to read through it. You can remove the custom type structs, pointer copies, indirection, etc, by using a _little_ repetition. 


```// SetComponentsStatus updates the pod status map component
func SetComponentsStatus(ctx context.Context, k k8s.Client, req ctrl.Request) error {
	var s lokiv1beta1.LokiStack
	if err := k.Get(ctx, req.NamespacedName, &s); err != nil {
		if apierrors.IsNotFound(err) {
			return nil
		}
		return kverrors.Wrap(err, "failed to lookup lokistack", "name", req.NamespacedName)
	}


	var err error
	s.Status.Components = lokiv1beta1.LokiStackComponentStatus{}
	if s.Status.Components.Compactor, err = appendPodStatus(ctx, k, manifests.CompactorName(s.Name), s.Name, s.Namespace); err != nil {
		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelCompactorComponent)
	}
	if s.Status.Components.Querier, err = appendPodStatus(ctx, k, manifests.QuerierName(s.Name), s.Name, s.Namespace); err != nil {
		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelQuerierComponent)
	}
	if s.Status.Components.Distributor, err = appendPodStatus(ctx, k, manifests.DistributorName(s.Name), s.Name, s.Namespace); err != nil {
		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelDistributorComponent)
	}
	if s.Status.Components.QueryFrontend, err = appendPodStatus(ctx, k, manifests.QueryFrontendName(s.Name), s.Name, s.Namespace); err != nil {
		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelQueryFrontendComponent)
	}
	if s.Status.Components.Ingester, err = appendPodStatus(ctx, k, manifests.IngesterName(s.Name), s.Name, s.Namespace); err != nil {
		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelIngesterComponent)
	}

	return k.Status().Update(ctx, &s, &client.UpdateOptions{})
}
```

This is a simpler implementation that I think is easier to read and maintain.

blockloop: You can probably shrink it and keep the loop if you make status.Components into a `map[string][]string`. Either way I think it should be rewritten.

ewolinetz: either ^ or change `status.Components` into a map. no need to swap between the two structures

controllers/internal/status/components.go
```diff
@@ -0,0 +1,66 @@
...
+		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelDistributorComponent)
+	}
+	if status.Components.QueryFrontend, err = appendPodStatus(ctx, k, manifests.LabelQueryFrontendComponent, s.Name, s.Namespace); err != nil {
+		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelQueryFrontendComponent)
+	}
+	if status.Components.Ingester, err = appendPodStatus(ctx, k, manifests.LabelIngesterComponent, s.Name, s.Namespace); err != nil {
+		return kverrors.Wrap(err, "failed lookup LokiStack component pods status", "name", manifests.LabelIngesterComponent)
+	}
+
+	s.Status = *status
```
blockloop, line 47: Since the scope of `s` is constrained to this func, you don't need to do a DeepCopy. You can just update s.State directly.

periklis: Good catch, I've addressed that using the `s.Status` directly.

periklis:
@blockloop PTAL

controllers/lokistack_controller.go
```diff
@@ -52,8 +53,17 @@ var (
...
-	deleteOnlyPred = builder.WithPredicates(predicate.Funcs{
-		UpdateFunc: func(e event.UpdateEvent) bool { return false },
+	updateOrDeleteOnlyPred = builder.WithPredicates(predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			switch e.ObjectOld.(type) {
+			case *appsv1.Deployment:
+			case *appsv1.StatefulSet:
+				return true
+			default:
+				return false
```
ewolinetz, line 63: nit: not sure we need this since we `return false` after the switch block

ewolinetz:
one nit otherwise lgtm

PR #51, opened 2021-05-11
LOG-1289: set pvc quotas per size

blockloop:


Makefile
```diff
@@ -109,12 +109,6 @@ manifests: $(CONTROLLER_GEN)
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 
-# Download golangci-lint locally if necessary
```
blockloop, line 112: This is cruft from before bingo was added.

ewolinetz (approved):

PR #53, opened 2021-05-14
Set lokistack condition failed/pending/ready based on pod status map

periklis:
### Description
The `LokiStack` resource represents a kubernetes resource for all owned sub-resources (e.g. Deployments, StatefulSets, ConfigMap, Services). Each of these resources runs through its own lifecycle expressed in their `Status.Conditions`. Missing/Non-fullfilled dependencies usually affect the running pods setting them to `Phase: Pending`. Hence, this PR provides translates the current pod status map into `LokiStack` conditions, i.e.:
1. If at least one pod is in `pending` the `LokiStack` CR is set to condition `Pending`.
2. If at least one pod is in `failed` the `LokiStack` CR is set to condition `Failed`.
3. In any other case the `LokiStack` CR is set to condition `Ready`.

#### List view
image

#### Details view
image


internal/status/status.go
```diff
@@ -0,0 +1,61 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// Refresh executes an aggregate update of the LokiStack Status struct, i.e.
+// - It recreates the Status.Components pod status map per component.
+// - It sets the appropriate Status.Condition to true that matches the pod status maps.
+func Refresh(ctx context.Context, k k8s.Client, req ctrl.Request) error {
+	if err := SetComponentsStatus(ctx, k, req); err != nil {
```
ewolinetz, line 19: side question, why is this not using the reference of req? wouldn't that save us a lot of needing to do deep copies of the status and then updating it? the req object would always reflect the most up-to-date and accurate state then...

periklis: AFAICS, the `DeepCopy` calls are all removed. The whole status pkg is designed to be self-sufficient in terms of requesting the `LokiStack` anew on public functions, e.g. `SetComponentsStatus`. Passing a reference to the LokiStack here makes error handling of situation like having an old version of `LokiStack` an issue for the calling site.

ewolinetz (approved):

PR #58, opened 2021-05-31
Log-1376: Add service monitor for ocp prometheus to scrape metrics

sasagarw:
This PR:
- Adds a service monitor for each loki component of loki cluster.
- Updates `MutateFuncFor` function to add support for service monitor type.
- Add unit test cases wherever applicable for the same.

/cc @periklis

Links:
Jira - https://issues.redhat.com/browse/LOG-1376


internal/manifests/var.go
```diff
@@ -28,6 +35,19 @@ const (
...
+	labelJobNameDistributor string = "monitor-loki-distributor"
+	// labelJobNameIngester is the label value for the ingester serviceMonitor job label
+	labelJobNameIngester string = "monitor-loki-ingester"
+	// labelJobNameQuerier is the label value for the querier serviceMonitor job label
+	labelJobNameQuerier string = "monitor-loki-querier"
+	// labelJobNameCompactor is the label value for the compactor serviceMonitor job label
+	labelJobNameCompactor string = "monitor-loki-compactor"
+	// labelJobNameQueryFrontend is the label value for the query-frontend serviceMonitor job label
+	labelJobNameQueryFrontend string = "monitor-loki-query-frontend"
+)
```
periklis, line 49: These constants are used as values for the ServiceMonitorSpec's `JobLabel` field. According to the docs:
```
❯ oc explain servicemonitor.spec.jobLabel
KIND:     ServiceMonitor
VERSION:  monitoring.coreos.com/v1

FIELD:    jobLabel <string>

DESCRIPTION:
     The label to use to retrieve the job name from.
```

`JobLabel` refers to the metrics label that should be used to identify the scraping job. AFAICS, we expose only the following labels on the services:
```
"app.kubernetes.io/name":     "loki",
"app.kubernetes.io/provider": "openshift",
"loki.grafana.com/name":       stackName,
"loki.grafana.com/component": component,
```

main.go
```diff
@@ -43,6 +44,8 @@ func init() {
 
 	utilruntime.Must(lokiv1beta1.AddToScheme(scheme))
 	// +kubebuilder:scaffold:scheme
+
+	utilruntime.Must(monitoringv1.AddToScheme(scheme))
```
periklis, line 48: Let's put a `TODO` here to make this loading optional via a `loki-operator` CLI flag.

periklis (changes requested):
Awesome headstart! In general it moves into the right direction and you can find below some improvment suggestions. Nevertheless a couple of key points are missing here:
1. The service resources per component miss the `service.beta.openshift.io/serving-cert-secret-name: serviceName` annotation to get the OCP cert service to create the PKI secrets.
2. Same Deployments/Statefulsets need to mount the generated secrets and expose the HTTP endpoint via HTTPS.

:rocket:

internal/manifests/var.go
```diff
@@ -13,6 +15,11 @@ const (
 
 	// DefaultContainerImage declares the default fallback for loki image.
 	DefaultContainerImage = "docker.io/grafana/loki:2.2.1"
+
+	// PrometheusCAFile declares the path for prometheus CA file for service monitors.
+	PrometheusCAFile string = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
```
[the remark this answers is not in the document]
sasagarw: Yeah I verified, there is no such constants provided by the library.

internal/manifests/mutate.go
```diff
@@ -82,3 +90,13 @@ func mutateStatefulSet(existing, desired *appsv1.StatefulSet) {
...
 	existing.Spec.VolumeClaimTemplates = desired.Spec.VolumeClaimTemplates
 }
+
+func mutateServiceMonitor(existing, desired *monitoringv1.ServiceMonitor) {
+	// ServiceMonitor selector is immutable so we set this value only if
+	// a new object is going to be created
+	if existing.CreationTimestamp.IsZero() {
+		existing.Spec.Selector = desired.Spec.Selector
+	}
+	existing.Spec.JobLabel = desired.Spec.JobLabel
```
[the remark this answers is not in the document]
sasagarw: Yes, it's static always. Have removed this.

internal/manifests/compactor.go
```diff
@@ -206,3 +209,33 @@ func NewCompactorHTTPService(opt Options) *corev1.Service {
 		},
 	}
 }
+
+// NewCompactorServiceMonitor creates a k8s service monitor for the compactor component
+func NewCompactorServiceMonitor(stackName, namespace string) *monitoringv1.ServiceMonitor {
+	l := ComponentLabels(LabelCompactorComponent, stackName)
+
+	serviceMonitorName := fmt.Sprintf("%s-%s-%s", labelJobNameCompactor, stackName, "cluster")
```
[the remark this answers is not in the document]
sasagarw: have removed it.

internal/manifests/mutate.go
```diff
@@ -82,3 +90,13 @@ func mutateStatefulSet(existing, desired *appsv1.StatefulSet) {
 	existing.Spec.Template = desired.Spec.Template
 	existing.Spec.VolumeClaimTemplates = desired.Spec.VolumeClaimTemplates
 }
+
+func mutateServiceMonitor(existing, desired *monitoringv1.ServiceMonitor) {
+	// ServiceMonitor selector is immutable so we set this value only if
+	// a new object is going to be created
+	if existing.CreationTimestamp.IsZero() {
+		existing.Spec.Selector = desired.Spec.Selector
+	}
```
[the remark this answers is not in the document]
sasagarw: As discussed by Eric and QE team, we don't want to let update anything as of now. Leaving it blank for now.

internal/manifests/mutate.go
```diff
@@ -82,3 +90,8 @@ func mutateStatefulSet(existing, desired *appsv1.StatefulSet) {
 	existing.Spec.Template = desired.Spec.Template
 	existing.Spec.VolumeClaimTemplates = desired.Spec.VolumeClaimTemplates
 }
+
+func mutateServiceMonitor(existing, desired *monitoringv1.ServiceMonitor) {
+	// ServiceMonitor selector is immutable so we set this value only if
+	// a new object is going to be created
+}
```
periklis, line 97: Does this imply that we don't support mutation of service monitor resources? Or is this a hindsight?

internal/manifests/var.go
```diff
@@ -113,6 +120,43 @@ func serviceNameCompactorHTTP(stackName string) string {
...
+}
+
 func fqdn(serviceName, namespace string) string {
 	return fmt.Sprintf("%s.%s.svc.cluster.local", serviceName, namespace)
 }
+
+// jobLabelForComponent is a servicemonitor.spec.jobLabel including the loki.grafana.com/component:<component> label
+func jobLabelForComponent(component string) string {
+	return fmt.Sprintf("loki.grafana.com/component: %s", component)
+}
```
periklis, line 138: The `ServiceMonitor.Spec.JobLabel` fields only needs the value of the label key, e.g. `loki.grafana.com/component`. Prometheus will pick the value by reading this key from the time series and use it as the scrape job name.

periklis (changes requested):
One small glitch and it LGTM.

internal/manifests/mutate.go
sasagarw: Yeah, it's kind of hindsight as of now.

periklis: Why? Is this going to stay empty?

sasagarw: Currently we don't want to provide any dynamic updation from users so we don't need to mutate. Hence it should be empty right?

periklis: As per now, we don't really manipulate anything in the ServiceMonitors, you are right. They are tightly coupled to a LokiStack name and thus not mutable. I tend to err that this can stay empty for now, as we might have to handle TLSConfig changes introduced by #59

sasagarw: So can we let this stay empty now? Once TLSConfig PR is merged, I will create another task in which we can work on the loki-operator CLI flag (`utilruntime.Must(monitoringv1.AddToScheme(scheme))`) optional and adding the TLSConfig changes.

internal/manifests/compactor_test.go
```diff
@@ -54,3 +54,13 @@ func TestNewCompactorStatefulSet_HasTemplateConfigHashAnnotation(t *testing.T) {
...
 }
+
+func TestNewCompactorServiceMonitor_SelectorMatchesLabels(t *testing.T) {
+	sts := manifests.NewCompactorServiceMonitor("abcd", "efgh")
+
+	l := sts.GetObjectMeta().GetLabels()
+	for key, value := range sts.Spec.Selector.MatchLabels {
+		require.Contains(t, l, key)
+		require.Equal(t, l[key], value)
+	}
```
periklis, line 65: This test is not addressing the core label matching issue with ServiceMonitors. The key thing to check here is that `ServiceMonitor.Spec.Selector.matchLabels` is matching the `Service.Spec.Selector`. This ensures that whenever we change the one we need to change the other to keep scraping working.

periklis (changes requested):

periklis (approved):

PR #59, opened 2021-06-02
Log 1403: OCP Certificate Signing

Red-GV:
This PR adds a flag for OpenShift deployment (`OCP_Deployment`) in order to enable OCP specific capabilities and an additional script which will create a secret based on the AWS configuration for development.

I believe the extra overlay from this PR should cover the OCP CSV: https://github.com/ViaQ/loki-operator/pull/55

JIRA: https://issues.redhat.com/browse/LOG-1403

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,9 +53,6 @@ BUNDLE_IMG ?= quay.io/$(REGISTRY_ORG)/loki-operator-bundle:$(VERSION)
 
 GO_FILES := $(shell find . -type f -name '*.go')
 
-# RUN_FLAGS defines the tags that are run with the operator.
-RUN_FLAGS ?= ""
-
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/$(REGISTRY_ORG)/loki-operator:$(VERSION)
 
@@ -88,7 +85,7 @@ manager: generate
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate manifests
-	go run ./main.go $(RUN_FLAGS)
+	go run ./main.go
 
 # Install CRDs into a cluster
 install: manifests $(KUSTOMIZE)
--- a/Makefile
+++ b/Makefile
@@ -174,7 +174,7 @@ endif
 # Build and push the secret for the S3 storage
 .PHONY: olm-deploy-example-storage-secret
 olm-deploy-example-storage-secret:
-	shell hack/deploy-example-secret.sh $(CLUSTER_LOGGING_NS)
+	hack/deploy-example-secret.sh $(CLUSTER_LOGGING_NS)
 
 .PHONY: olm-deploy-example
 olm-deploy-example: olm-deploy olm-deploy-example-storage-secret
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,11 @@ BUNDLE_IMG ?= quay.io/$(REGISTRY_ORG)/loki-operator-bundle:$(VERSION)
 
 GO_FILES := $(shell find . -type f -name '*.go')
 
+# GO_BUILD_TAGS defines the tags to build the operator binary. The outside of the
+# default, the only other option is "openshift". Use this in order to enable, the
+# usage of Openshift specific features.
+GO_BUILD_TAGS ?= ""
+
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/$(REGISTRY_ORG)/loki-operator:$(VERSION)
 
@@ -70,8 +75,6 @@ all: generate lint manager bin/loki-broker
 
 OCI_RUNTIME ?= $(shell which podman || which docker)
 
-OCP_DEPLOYMENT ?= false
-
 # Run tests
 test: generate go-generate lint manifests
 test: $(GO_FILES)
@@ -83,7 +86,7 @@ scorecard: generate go-generate bundle
 
 # Build manager binary
 manager: generate
-	go build -o bin/manager main.go
+	go build -tags $(GO_BUILD_TAGS) -o bin/manager main.go
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate manifests
@@ -132,7 +135,6 @@ generate: $(CONTROLLER_GEN)
 
 # Build the image
 oci-build:
-	sed -i "s/\tIsOCPDeployment.*/\tIsOCPDeployment bool = ${OCP_DEPLOYMENT}/" internal/manifests/options.go
 	$(OCI_RUNTIME) build -t ${IMG} .
 
 # Push the image
@@ -161,11 +163,6 @@ olm-deploy-bundle: bundle bundle-build
 .PHONY: olm-deploy-operator
 olm-deploy-operator: oci-build oci-push
 
-# Build and push the secret for the S3 storage
-.PHONY: olm-deploy-example-storage-secret
-olm-deploy-example-storage-secret:
-	sh hack/deploy-example-secret.sh $(CLUSTER_LOGGING_NS)
-
 # Deploy the operator bundle and the operator via OLM into
 # an Kubernetes cluster selected via KUBECONFIG.
 .PHONY: olm-deploy
@@ -179,9 +176,14 @@ olm-deploy: olm-deploy-bundle olm-deploy-operator $(OPERATOR_SDK)
 	$(OPERATOR_SDK) run bundle -n $(CLUSTER_LOGGING_NS) --install-mode OwnNamespace $(BUNDLE_IMG)
 endif
 
+# Build and push the secret for the S3 storage
+.PHONY: olm-deploy-example-storage-secret
+olm-deploy-example-storage-secret:
+	shell hack/deploy-example-secret.sh $(CLUSTER_LOGGING_NS)
+
 .PHONY: olm-deploy-example
 olm-deploy-example: olm-deploy olm-deploy-example-storage-secret
-	kubectl -n $(CLUSTER_LOGGING_NS) create -f config/samples/loki_v1beta1_lokistack.yaml
+	kubectl -n $(CLUSTER_LOGGING_NS) create -f hack/lokistack_dev.yaml
 
 # Cleanup deployments of the operator bundle and the operator via OLM
 # on an OpenShift cluster selected via KUBECONFIG.
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,8 @@ all: generate lint manager bin/loki-broker
 
 OCI_RUNTIME ?= $(shell which podman || which docker)
 
+OCP_DEPLOYMENT ?= false
+
 # Run tests
 test: generate go-generate lint manifests
 test: $(GO_FILES)
@@ -130,6 +132,7 @@ generate: $(CONTROLLER_GEN)
 
 # Build the image
 oci-build:
+	sed -i "s/\tIsOCPDeployment.*/\tIsOCPDeployment bool = ${OCP_DEPLOYMENT}/" internal/manifests/options.go
 	$(OCI_RUNTIME) build -t ${IMG} .
 
 # Push the image
@@ -158,6 +161,11 @@ olm-deploy-bundle: bundle bundle-build
 .PHONY: olm-deploy-operator
 olm-deploy-operator: oci-build oci-push
 
+# Build and push the secret for the S3 storage
+.PHONY: olm-deploy-example-storage-secret
+olm-deploy-example-storage-secret:
+	sh hack/deploy-example-secret.sh $(CLUSTER_LOGGING_NS)
+
 # Deploy the operator bundle and the operator via OLM into
 # an Kubernetes cluster selected via KUBECONFIG.
 .PHONY: olm-deploy
@@ -170,6 +178,10 @@ olm-deploy: olm-deploy-bundle olm-deploy-operator $(OPERATOR_SDK)
 	kubectl label ns/$(CLUSTER_LOGGING_NS) openshift.io/cluster-monitoring=true --overwrite
 	$(OPERATOR_SDK) run bundle -n $(CLUSTER_LOGGING_NS) --install-mode OwnNamespace $(BUNDLE_IMG)
 endif
+
+.PHONY: olm-deploy-example
+olm-deploy-example: olm-deploy olm-deploy-example-storage-secret
+	kubectl -n $(CLUSTER_LOGGING_NS) create -f config/samples/loki_v1beta1_lokistack.yaml
 
 # Cleanup deployments of the operator bundle and the operator via OLM
 # on an OpenShift cluster selected via KUBECONFIG.
--- a/Makefile
+++ b/Makefile
@@ -53,10 +53,8 @@ BUNDLE_IMG ?= quay.io/$(REGISTRY_ORG)/loki-operator-bundle:$(VERSION)
 
 GO_FILES := $(shell find . -type f -name '*.go')
 
-# GO_BUILD_TAGS defines the tags to build the operator binary. The outside of the
-# default, the only other option is "openshift". Use this in order to enable, the
-# usage of Openshift specific features.
-GO_BUILD_TAGS ?= ""
+# RUN_FLAGS defines the tags that are run with the operator.
+RUN_FLAGS ?= ""
 
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/$(REGISTRY_ORG)/loki-operator:$(VERSION)
@@ -86,11 +84,11 @@ scorecard: generate go-generate bundle
 
 # Build manager binary
 manager: generate
-	go build -tags $(GO_BUILD_TAGS) -o bin/manager main.go
+	go build -o bin/manager main.go
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate manifests
-	go run ./main.go
+	go run ./main.go $(RUN_FLAGS)
 
 # Install CRDs into a cluster
 install: manifests $(KUSTOMIZE)
--- a/config/overlays/openshift/manager_run_flags_patch.yaml
+++ b/config/overlays/openshift/manager_run_flags_patch.yaml
@@ -10,6 +10,6 @@ spec:
       containers:
         - name: manager
           args:
-          - "--with-ocp-features"
-          - "--with-service-monitors"
           - "--with-cert-signing-service"
+          - "--with-service-monitors"
+          - "--with-tls-service-monitors"
--- a/config/overlays/openshift/manager_run_flags_patch.yaml
+++ b/config/overlays/openshift/manager_run_flags_patch.yaml
@@ -8,9 +8,8 @@ spec:
   template:
     spec:
       containers:
-        - command: ["go run"]
-          name: manager
+        - name: manager
           args:
-          - "--with-ocp-mode"
+          - "--with-ocp-features"
           - "--with-service-monitors"
           - "--with-tls-service-monitors"
--- a/config/overlays/openshift/manager_run_flags_patch.yaml
+++ b/config/overlays/openshift/manager_run_flags_patch.yaml
@@ -12,4 +12,4 @@ spec:
           args:
           - "--with-ocp-features"
           - "--with-service-monitors"
-          - "--with-tls-service-monitors"
+          - "--with-cert-signing-service"
--- a/config/overlays/openshift/manager_run_flags_patch.yaml
+++ b/config/overlays/openshift/manager_run_flags_patch.yaml
@@ -1,0 +1,16 @@
+# This patch inject a sidecar container which is a HTTP proxy for the
+# controller manager, it performs RBAC authorization against the Kubernetes API using SubjectAccessReviews.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+spec:
+  template:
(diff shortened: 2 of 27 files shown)


periklis (changes requested):
Awesome work! Keep :rocket: !

Couple of improvements annotated along the source code, please take a look. Further on, the following points are missing:
1. Mounting secrets and using the ca/cert/key triplet by loki containers.
2. CLIs flags for `--with-ocp-mode` and `--with-tls-service-monitor-config`

Makefile
```diff
@@ -158,6 +158,11 @@ olm-deploy-bundle: bundle bundle-build
 .PHONY: olm-deploy-operator
 olm-deploy-operator: oci-build oci-push
 
+# Build and push the secret for the S3 storage
+.PHONY: olm-deploy-example-storage-secret
+olm-deploy-example-storage-secret:
+	sh hack/deploy-example-secret.sh $(CLUSTER_LOGGING_NS)
```
periklis, line 164: nit. No need to prepend `sh` here, as Makefile's `SHELL` will work for that fine.

hack/deploy-example-secret.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
```
periklis, line 2: Recommendation to use `set -eou pipefail` by default. This helps a lot with pipes and errors among other things.

hack/lokistack_dev.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: loki.openshift.io/v1beta1
+kind: LokiStack
+metadata:
+  name: lokistack-dev
+spec:
+  size: 1x.extra-small
+  replicationFactor: 2
```
periklis, line 7: The replication factor for `1x.extra-small` can be only 1 for now as it supports a single ingester instance. In turn this helps to restart the ingester faster when testing dev changes.

Makefile
```diff
@@ -81,7 +81,7 @@ scorecard: generate go-generate bundle
 
 # Build manager binary
 manager: generate
-	go build -o bin/manager main.go
+	go build -o bin/manager main.go -tags openshift
```
periklis, line 84: Appending the tag `openshift` should be an optional Makefile variable. By default we should not add the flag but Makefile users should be able to do something like `make ... DISTRIBUTION=openshift`

periklis:
> Awesome work! Keep rocket !
> 
> Couple of improvements annotated along the source code, please take a look. Further on, the following points are missing:
> 
>     1. Mounting secrets and using the ca/cert/key triplet by loki containers.
> 
>     2. CLIs flags for `--with-ocp-mode` and `--with-tls-service-monitor-config`

@Red-GV Comments on the :point_up:

main.go
```diff
@@ -43,26 +44,37 @@ func init() {
...
+		"Enables features in an Openshift cluster.")
+	flag.BoolVar(&useTLSServiceMonitorConfig, "with-tls-service-monitor-config", false,
+		"Enables loading of a prometheus service monitor.")
 	flag.Parse()
 
 	log.Init("loki-operator")
 	ctrl.SetLogger(log.GetLogger())
 
+	if useTLSServiceMonitorConfig {
+		// +kubebuilder:scaffold:scheme
```
periklis, line 70: This annotation should stay in the `init` function to keep operator-sdk working.

main.go
```diff
@@ -43,26 +44,37 @@ func init() {
...
+		utilruntime.Must(monitoringv1.AddToScheme(scheme))
```
periklis, line 71: Actually this should be guarded by a `--with-service-monitors` flag. The `--with-tls-service-monitor-config` belongs to the servicemonitor endpoint builder in `manifests/var.go`. With TLS enabled:
```go
// serviceMonitorLokiEndPoint returns the loki endpoint for service monitors.
func serviceMonitorLokiEndPoint(stackName, serviceName, namespace string) monitoringv1.Endpoint {
	tlsConfig := serviceMonitorTLSConfig(serviceName, namespace)

	return monitoringv1.Endpoint{
		Port:            stackName,
		Path:            "/metrics",
		Scheme:          "https",
		BearerTokenFile: BearerTokenFile,
		TLSConfig:       &tlsConfig,
	}
}
```
Without TLS:
```go
// serviceMonitorLokiEndPoint returns the loki endpoint for service monitors.
func serviceMonitorLokiEndPoint(stackName, serviceName, namespace string) monitoringv1.Endpoint {
	return monitoringv1.Endpoint{
		Port:            stackName,
		Path:            "/metrics",
		Scheme:          "http",
	}
}
```
Both should be unified in one function using `useTLSServiceMonitorConfig`

periklis:
@Red-GV Regarding mounting the secrets, since you use the OCP cert signing service which creates the secret for you, you don't need to add any manual certificate generation. If a secret with a valid PKI is available from the OCP cert signing service, we can use this secret in a loki component container by:
1. Mount the secret as a volume
2. Use the ca/cert/key provided by the secret via Loki's cli flags like this:
```
-server.http-tls-ca-path
-server.http-tls-cert-path
-server.http-tls-key-path
```

Red-GV:
@periklis I haven't cleaned up or tested this code yet, but I did make an attempt adding the volume mounts (code structure wise) for a style review.

config/overlays/openshift/manager_run_flags_patch.yaml
```diff
@@ -0,0 +1,16 @@
...
+# controller manager, it performs RBAC authorization against the Kubernetes API using SubjectAccessReviews.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+spec:
+  template:
+    spec:
+      containers:
+        - command: ["go run"]
```
periklis, line 11: Why `go run` and not the binary directly? As per go binaries being selfsustained we don't need to load the golang runtime into target containers and thus squeeze them in total size.

internal/handlers/lokistack_create_or_update.go
```diff
@@ -63,11 +63,13 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
-		Image:         img,
-		Stack:         stack.Spec,
-		ObjectStorage: *storage,
+		Name:                        req.Name,
+		Namespace:                   req.Namespace,
+		Image:                       img,
+		Stack:                       stack.Spec,
+		ObjectStorage:               *storage,
+		UseServiceMonitors:          manifests.UseServiceMonitors,
+		UseTLSEnabledServiceMonitor: manifests.UseTLSEnabledServiceMonitors,
```
periklis, line 72: nit. rename to `UseTLSServiceMonitorConfig` as it represents better the usage of a `TLSConfig` for Service Monitor endpoints.

periklis, line 72: nit. Suggest to switch from `Use` to `Enable` as a prefix, as we literally speak of enabling a feature via this flag.

internal/manifests/var.go
```diff
@@ -131,10 +141,38 @@ func serviceNameQueryFrontendHTTP(stackName string) string {
 	return fmt.Sprintf("loki-query-frontend-http-%s", stackName)
 }
 
+func signingServiceSecretName(stackName string) string {
+	return fmt.Sprintf("%s-%s", stackName, "metrics")
+}
```
periklis, line 146: :x:  The implementation here makes a wrong assumption namely that secret's name is based on `stackName` and a suffix `-metrics`. However the cert signing service will create one secret per Loki component service, e.g.
- Distributor service: `loki-distributor-http-my-stack-name` --> Secret: `loki-distributor-http-my-stack-name-metrics`
- Ingester service: `loki-ingester-http-my-stack-name` --> Secret: `loki-ingester-http-my-stack-name-metrics`

The approarch to take here is to pass the service name per component here.

internal/manifests/var.go
```diff
@@ -131,10 +141,38 @@ func serviceNameQueryFrontendHTTP(stackName string) string {
...
+
 func fqdn(serviceName, namespace string) string {
 	return fmt.Sprintf("%s.%s.svc.cluster.local", serviceName, namespace)
 }
 
+func volumeMounts(podSpec corev1.PodSpec, stackName string) corev1.podSpec {
```
periklis, line 152: :x: The function's name is definetly not representative of the work that it accomplishes namely mounting volumes, extending container arguments.

:x: In addition the function is dedicated to patch the pod spec for our service secret, however it does not register the volumes which get mounted to the pod spec.

internal/manifests/var.go
```diff
@@ -131,10 +141,38 @@ func serviceNameQueryFrontendHTTP(stackName string) string {
...
+	volumeMounts := adjusted.Containers[0].VolumeMounts
+
+	volumeMounts = append(volumeMounts, corev1.VolumeMount{
+		Name:      signingServiceSecretName(stackName),
+		MountPath: "/etc/proxy/secrets",
+	})
+	volumeMounts = append(volumeMounts,corev1.VolumeMount{
+		Name:      "certificates",
+		MountPath: "/etc/proxy/loki",
+	})
```
periklis, line 165: :x: Why is the purpose of two volumenMounts? FWIW the cert signing service is creating only one secret which translates in one Volume registration in the pod spec and one volume mount into the Loki component container.

main.go
```diff
@@ -44,25 +45,36 @@ func init() {
...
-	// TODO: make this loading optional via a loki-operator CLI flag.
-	utilruntime.Must(monitoringv1.AddToScheme(scheme))
 }
 
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+	var enableServiceMonitors bool
+	var useTLSServiceMonitors bool
```
periklis, line 55: :bulb: Let's be consistent here and use `enable` as a prefix for enabling feature flags.

main.go
```diff
@@ -44,25 +45,36 @@ func init() {
...
+	flag.BoolVar(&useTLSServiceMonitors, "with-tls-service-monitors", false,
+		"Enables loading of a prometheus service monitor.")
 	flag.Parse()
 
 	log.Init("loki-operator")
 	ctrl.SetLogger(log.GetLogger())
 
+	if enableServiceMonitors || useTLSServiceMonitors {
+		manifests.UseServiceMonitors = true
+		manifests.UseTLSEnabledServiceMonitors = useTLSServiceMonitors
```
periklis, line 73: :grey_question: Any chance to avoid using global variables for these feature flags?

periklis (changes requested):
:rocket: Super awesome progress! Keep that :runner: !

In general this goes into the right direction, couple of improvements and two general asks here:
1. Could you elaborate on why not using `mergo.Merge` for merging manifest patches? Which pitfalls did you encounter?
2. The new patch `config/overlays/openshift` is not relfected into the target bundle. Can you put a reminder to add the up2date contents after a `make bundle` run?

Red-GV:
I had not started yet with the `mergo` because I wanted to a non-error producing version first. Though I'm somewhat concerned that using the overwrite will just overwrite the `Volumes` and other arrays as a result. I'm experimenting with the `AppendSlice` but going down that road might just be the same result with an possibility to produce an error.

I also need to get the CA bundle from the service and have it put it into the secret. I've only added the addition of that cert and what not to the `distributor.go`. After finishing with that experimentation with `mergo`, then I'll do it to the other components.

internal/manifests/mutate.go
```diff
@@ -95,3 +96,56 @@ func mutateServiceMonitor(existing, desired *monitoringv1.ServiceMonitor) {
 	// ServiceMonitor selector is immutable so we set this value only if
 	// a new object is going to be created
 }
+
+func mutatePodSpecForTLSEnablement(podSpec *corev1.PodSpec, serviceName string) {
```
periklis, line 100: This function does not belong here. It is augmenting a pod spec if a feature flag is enabled. I tend to err this belongs more to `service_monitors.go`. In addition it is literarly configuring the pod spec to use the generated PKI. WDYT of calling such functions `configureServiceMonitorPKI()`.

This would look on the caller site like:
```go
if opt.EnableTLSServiceMonitors {
    configureServiceMonitorPKI(podSpec, serviceName)
}

File: internal/manifests/mutate.go [hunk 15]
```diff
@@ -95,3 +96,56 @@ func mutateServiceMonitor(existing, desired *monitoringv1.ServiceMonitor) {
...
+
+	// if err := mergo.Merge(podSpec.Volumes, tlsSpec.Volumes, mergo.WithAppendSlice); err != nil {
+	// 	return kverrors.Wrap(err, "failed to merge volumes")
+	// }
+	//
+	// if err := mergo.Merge(podSpec.Containers[0], tlsSpec.Containers[0], mergo.WithAppendSlice); err != nil {
+	// 	return kverrors.Wrap(err, "failed to merge containers")
+	// }
+	//
+	// return nil
```
periklis, line 150: Isn't this form more readable than the above mutations? I feel yes.

internal/handlers/lokistack_create_or_update.go
```diff
@@ -63,13 +63,13 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
...
-		ObjectStorage:               *storage,
-		UseServiceMonitors:          manifests.UseServiceMonitors,
-		UseTLSEnabledServiceMonitor: manifests.UseTLSEnabledServiceMonitors,
+		Name:                          req.Name,
+		Namespace:                     req.Namespace,
+		Image:                         img,
+		Stack:                         stack.Spec,
+		ObjectStorage:                 *storage,
+		EnableServiceMonitors:         manifests.UseServiceMonitors,
+		EnableTLSServiceMonitorConfig: manifests.UseTLSEnabledServiceMonitors,
```
periklis, line 72: Would you mind to also rename the right hand side to `Enable..` instead of `Use...`

periklis (changes requested):
Looks really awesome, I feel we are on the last mile with this one. Couple of improvements along the source code suggested. Besides two major points to attack:
1. After comparing inline mutation vs. mergo, I feel we should go with mergo and account for some extra error handling in our builders.
2. We need to add the volume mount / arg use on all components and not only on the distributor (Just a reminder).

internal/handlers/lokistack_create_or_update.go
Red-GV: Yeah, I haven't touched these ones yet as I'm still looking at a way of nicely accomplishing this: https://github.com/ViaQ/loki-operator/pull/59#discussion_r649726800

Red-GV:
Okay, doing one more tiny push for style review before I write this across the field. I'm kinda thinking that adding in the error into the build (which is probably the best place for them considering the naming), but it might make the `BuildAll` function more "ugly" - for lack of a better term. Not sure if you'll see the same.

Also, is the CA bundle a part of this card? I noticed that you mentioned that here https://github.com/ViaQ/loki-operator/pull/59#issuecomment-858329936. I think I'm going to need to make a new ConfigMap or something in order to get the bundle from the service (https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html#add-service-serving).

periklis:
> Okay, doing one more tiny push for style review before I write this across the field. I'm kinda thinking that adding in the error into the build (which is probably the best place for them considering the naming), but it might make the `BuildAll` function more "ugly" - for lack of a better term. Not sure if you'll see the same.

It is ok if we pay some error handling in `BuildAll`. We had in the early days some kind of error handling there which became obsolete, that's why it look cleaner today. Nevertheless the way it looks now is quite obvious and readable. Entrypoint builder methods like `BuildDistributor` build the initial version of the manifests whitout error effects. Next configure methods take the advantage of `mergo.Merge` to make the patches obvious to read as if they are manifests, but with some error effects, which should be very rare.

> 
> Also, is the CA bundle a part of this card? I noticed that you mentioned that here [#59 (comment)](https://github.com/ViaQ/loki-operator/pull/59#issuecomment-858329936). I think I'm going to need to make a new ConfigMap or something in order to get the bundle from the service (https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html#add-service-serving).

No actually not, we don't need to do the CA bundle part at all. This is more of my misunderstanding. Mounting the certificate and the key provided in the cert-signing service secret is enough. Prometheus has the CA bundle itself to verify that our certs are good to use.

internal/manifests/compactor.go
```diff
@@ -17,11 +17,10 @@ import (
 )
 
 // BuildCompactor builds the k8s objects required to run Loki Compactor.
-func BuildCompactor(opts Options) []client.Object {
-	return []client.Object{
-		NewCompactorStatefulSet(opts),
-		NewCompactorGRPCService(opts.Name),
-		NewCompactorHTTPService(opts.Name),
+func BuildCompactor(opts Options) (*appsv1.StatefulSet, []client.Object) {
```
periklis, line 20: Let's not break the signature here, please move the configure in here and make this function like `func BuildCompactor(opts Options) ([]client.Object, error)`. It should be an all or nothing builder.

internal/manifests/openshift.go
```diff
@@ -1,10 +1,8 @@
...
-var (
-	// UseCertificateSigningService test
-	UseCertificateSigningService = false
-	// UseServiceMonitors test
-	UseServiceMonitors = false
-	// UseTLSEnabledServiceMonitors test
-	UseTLSEnabledServiceMonitors = false
-)
+// OpenshiftOptions contains flags that activate Openshift features
+type OpenshiftOptions struct {
```
periklis, line 4: These Options are **not** Openshift-only related. Prefer to call them feature flags.

main.go
```diff
@@ -51,14 +51,15 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+	var enableCertSigning bool
 	var enableServiceMonitors bool
 	var enableTLSServiceMonitors bool
```
periklis: One `var(..)` block to rule them all and improve reability.

internal/manifests/compactor.go
Red-GV: Sorry, misunderstood. Should be good now.

internal/manifests/openshift.go
```diff
@@ -0,0 +1 @@
+package manifests
```
periklis, line 1: Do we need this file at all?

main.go
```diff
@@ -44,25 +45,37 @@ func init() {
...
+		enableTLSServiceMonitors bool
+	)
+
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&enableCertSigning, "with-ocp-features", false,
+		"Enables features in an Openshift cluster.")
```
periklis, line 66: `with-ocp-features` -> `with-cert-signing-service`

periklis (changes requested):

config/overlays/openshift/manager_run_flags_patch.yaml
```diff
@@ -12,4 +12,4 @@ spec:
           args:
           - "--with-ocp-features"
           - "--with-service-monitors"
-          - "--with-tls-service-monitors"
+          - "--with-cert-signing-service"
```
periklis: Looks like this change hit the wrong flag. The list should look like this:
```
          - "--with-service-monitors"
          - "--with-tls-service-monitors"
          - "--with-cert-signing-service"
```

internal/manifests/build_test.go
```diff
@@ -81,41 +81,50 @@ func TestApplyUserOptions_AlwaysSetCompactorReplicasToOne(t *testing.T) {
...
+					EnableTLSServiceMonitorConfig:   false,
+				},
+			},
 		},
 	}
 
-	for index, tst := range table {
-		testName := fmt.Sprintf("%s_service_monitor_count_%v", tst.BuildOptions.Name, index)
+	for _, tst := range table {
+		testName := fmt.Sprint(tst.desc)
```
periklis, line 127: No need for `fmt.Sprint` anymore, a simple `tst := tst` is simply enough

internal/manifests/build_test.go
```diff
@@ -130,50 +139,59 @@ func TestBuildAll_DidBuildServiceMonitors(t *testing.T) {
...
```
periklis, line 182: No need for `fmt.Sprint` anymore, a simple `tst := tst` is simply enough

periklis (changes requested):

internal/manifests/build_test.go
```diff
@@ -124,8 +123,7 @@ func TestBuildAll_WithFeatureFlags_EnableServiceMonitors(t *testing.T) {
 	}
 
 	for _, tst := range table {
-		testName := fmt.Sprint(tst.desc)
-		t.Run(testName, func(t *testing.T) {
+		t.Run(tst.desc, func(t *testing.T) {
```
periklis, line 126: Encourage you to use `tst := tst` before `t.Run` since you are using `t.Parallel()` and for-loop. Look here:
https://github.com/golang/go/wiki/CommonMistakes

periklis (approved):

PR #63, opened 2021-06-29
Add OpenShift cluster storage support

vyzigold:
This PR adds support for using OCS for S3 storage.

Object bucket claim can be created with config/overlays/development/ocs, the manager then reads the generated secret and config map created by that claim and sets the required options in the loki config file similarly to how minio secret is used.

I tested with OCP 4.6.32, OCS 4.6.5 with insecure_skip_verify set to true. (I didn't have the proper certificates configured for https veryfication.)


vyzigold:
I wanted to try deploying with OCS and thought it might be useful to you too. This is my first encounter with a go-based operator, so I fully expect to move some stuff around.

PR #69, opened 2021-08-15
add reconcile for dependent objects ( such as the storage secret )

eranra:
When changing one of the dependent objects, specifically storage secret the operator need to redeploy relevant objects

### DIFF
--- a/controllers/lokistack_controller.go
+++ b/controllers/lokistack_controller.go
@@ -20,6 +20,10 @@ import (
 	"context"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
 	"github.com/ViaQ/loki-operator/controllers/internal/management/state"
 	"github.com/ViaQ/loki-operator/internal/external/k8s"
 	"github.com/ViaQ/loki-operator/internal/handlers"
@@ -76,9 +80,10 @@ var (
 // LokiStackReconciler reconciles a LokiStack object
 type LokiStackReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
-	Flags  manifests.FeatureFlags
+	Log                 logr.Logger
+	Scheme              *runtime.Scheme
+	Flags               manifests.FeatureFlags
+	DependentObjectsMap map[types.NamespacedName]types.NamespacedName
 }
 
 // +kubebuilder:rbac:groups=loki.openshift.io,resources=lokistacks,verbs=get;list;watch;create;update;patch;delete
@@ -98,7 +103,13 @@ type LokiStackReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/reconcile
-func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *LokiStackReconciler) Reconcile(ctx context.Context, orgReq ctrl.Request) (ctrl.Result, error) {
+	// Initiate reconcile relevant loki stack from dependent objects (e.g. secrets, configmaps)
+	req := orgReq
+	if val, ok := r.DependentObjectsMap[req.NamespacedName]; ok {
+		req = ctrl.Request{NamespacedName: val}
+	}
+
 	ok, err := state.IsManaged(ctx, req, r.Client)
 	if err != nil {
 		return ctrl.Result{
@@ -112,7 +123,7 @@ func (r *LokiStackReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
-	err = handlers.CreateOrUpdateLokiStack(ctx, req, r.Client, r.Scheme, r.Flags)
+	err = handlers.CreateOrUpdateLokiStack(ctx, req, r.Client, r.Scheme, r.Flags, r.DependentObjectsMap)
 	if err != nil {
 		return ctrl.Result{
 			Requeue:      true,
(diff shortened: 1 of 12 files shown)


controllers/lokistack_controller.go
```diff
@@ -76,9 +80,10 @@ var (
 // LokiStackReconciler reconciles a LokiStack object
 type LokiStackReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
-	Flags  manifests.FeatureFlags
+	Log                 logr.Logger
+	Scheme              *runtime.Scheme
+	Flags               manifests.FeatureFlags
+	DependentObjectsMap map[types.NamespacedName]types.NamespacedName
```
periklis, line 86: Why do we need a map in the first place? Adding the watch directive below should be enough to reconcile on secret changes. If the purpose is to filter secret create/update/delete events, you can certainly do this with a simple predicate rather than maintaining a custom map.

internal/handlers/lokistack_create_or_update.go
```diff
@@ -61,6 +63,13 @@ func CreateOrUpdateLokiStack(ctx context.Context, req ctrl.Request, k k8s.Client
 		)
 	}
 
+	sha := sha1.New()
+	sha.Write([]byte(fmt.Sprintf("%v", s3secret)))
+	sha1Storage := fmt.Sprintf("%x", sha.Sum(nil))
```
periklis, line 68: This looks a bit orphaned here. Reading through this function makes me wonder about this low level details. I believe the sha1 should be a extra return value of `secrets.Extract` above.

internal/manifests/mutate.go
```diff
@@ -59,24 +66,31 @@ func MutateFuncFor(existing, desired client.Object) controllerutil.MutateFn {
 	}
 }
 
+func mergeWithOverride(dst, src interface{}) {
+	err := mergo.Merge(dst, src, mergo.WithOverride)
+	if err != nil {
+		panic(err)
```
periklis, line 72: We should certainly not panic here. We should handle errors properly.

internal/manifests/var.go
```diff
@@ -38,12 +38,40 @@ const (
...
+			for _, v2 := range res {
+				if v == v2 {
+					continue srcMap
+				}
+			}
+			res[k] = v
+		}
+	}
+	return res
+}
```
periklis, line 61: What is this function doing differently than `mergo.Merge`?

periklis (changes requested):
Thanks for submitting this for the storage secret! We certainly have a reload mechanism for changes in the configmaps, if something is broken there, we should address the configmap stuff in a separate PR.

Regarding the PR content, it looks to me that we have many convoluted changes that don't clearly belong to the purpose of this PR, e.g. changes in config values, missing ports, or PVC volumeMode. I would like to see these changes discussed and proposed in a separate PR.

eranra:
> Regarding the PR content, it looks to me that we have many convoluted changes that don't clearly belong to the purpose of this PR, e.g. changes in config values, missing ports, or PVC volumeMode. I would like to see these changes discussed and proposed in a separate PR.

All of those missing manifest entries are required for clean and correct reconcile processing. Before adding all those changes, every reconcile caused all of the stack objects to be updated ... so the comparison always failed. This is sure to the fact that k8s sets some default values in those fields and if they are not set the comparison function (even after moving the  mergo.Merge ) consider those diffret

controllers/lokistack_controller.go
eranra: If you look here:: https://github.com/ViaQ/loki-operator/pull/69/files#diff-5637f01ba13687b5d23f5d331aa3a063e9f3853d3dc7b1dec3911cc8d80315e8R71

The idea is that we can create a mapping between a dependent object and between some known object in lokistack. we recognize this during the first reconsile cycle. Currently, we have only one "example" of that but going forward I see more of that needed.  I will give an example - for the ruler we will want config map with rules ... this needs to redeploy just the ruler object(/s) but not the rest of the stack.

internal/handlers/lokistack_create_or_update.go
eranra: moved to `secrets.Extract`

internal/manifests/mutate.go
eranra: fixed

controllers/lokistack_controller.go
periklis: We don't need a dependency map to solve the restart issue at hand for now. If you use the watch + predicate of controller runtime, you can restart the cluster on secret events. Regarding smart restarts (e.g. ruler only), we spiked that before and we withdrawn the idea in #47. This is not a thing we want to have right now, without understanding loki component upgrades over loki versions. We skipped this for later.

periklis:
> > Regarding the PR content, it looks to me that we have many convoluted changes that don't clearly belong to the purpose of this PR, e.g. changes in config values, missing ports, or PVC volumeMode. I would like to see these changes discussed and proposed in a separate PR.
> 
> All of those missing manifest entries are required for clean and correct reconcile processing. Before adding all those changes, every reconcile caused all of the stack objects to be updated ... so the comparison always failed. This is sure to the fact that k8s sets some default values in those fields and if they are not set the comparison function (even after moving the mergo.Merge ) consider those diffret

That's  good thing then. Nevertheless it is not required for this PR and I advice to put this on a separate PR please. Work gets squashed in the end and we want to keep some sanity in the PR commits landing on master.

internal/manifests/var.go
eranra: Code changed to use mergo.Merge

controllers/lokistack_controller.go
eranra: @periklis I get it ... what I created is a very-very simple map ... actually one of the problems we will have if we drop that structure is that  the reconcile loop handles multiple lokistacks so we need to know which stack we need to reconcile on change to what secret ... I can't answer this question without that map ... 

@periklis Do you have an idea how to solve that ???

eranra:
> > > Regarding the PR content, it looks to me that we have many convoluted changes that don't clearly belong to the purpose of this PR, e.g. changes in config values, missing ports, or PVC volumeMode. I would like to see these changes discussed and proposed in a separate PR.
> > 
> > 
> > All of those missing manifest entries are required for clean and correct reconcile processing. Before adding all those changes, every reconcile caused all of the stack objects to be updated ... so the comparison always failed. This is sure to the fact that k8s sets some default values in those fields and if they are not set the comparison function (even after moving the mergo.Merge ) consider those diffret
> 
> That's good thing then. Nevertheless it is not required for this PR and I advice to put this on a separate PR please. Work gets squashed in the end and we want to keep some sanity in the PR commits landing on master.

@periklis OK, make sense ... I created another PR https://github.com/ViaQ/loki-operator/pull/72 with just the changes to the manifests and the merge function so that the reconsile will work as expected ... once you merge PR 72. I will rebase on top of that. Hope this works for you

PR #70, opened 2021-08-17
sanity-test

eranra:


periklis:
Thanks for providing a bit more sanity testing into this project. This looks like a smoke/integration test for loki. Although it looks pretty good, we should pursue for this one with a different mindset:
1. Such a job taking multiple minutes should not belong into the GH workflow machinery. We should aim for fast PR feedback.
2. Deploying dependencies for development/testing should not be handled manually in crafty bash code. These belong to the development [overlay](https://github.com/ViaQ/loki-operator/tree/master/config/overlays/development), e.g. minio already exists there.
3. We should aim for test code homogenity from the scratch and as per this projects test base is in golang (golang tests+testify) we should stick for it.
4. Integration testing as aimed with this PR is still missing the lokistack-gateway component, which is currently actively developed. Having the gateway in place we can follow up on using [observatorium-up](https://github.com/observatorium/up) for such tests.

eranra:
(1) @periklis I had a discussion on the time it takes to run this test with @Red-GV . If you take a look at the timing you will see that the amount of time it takes for this sanity test to run is equivalent to the time it takes for the scoreboard test to run so this is ** none-issue**. We are not talking about a full-blown E2E test. This is a sanity test to verify that things are working as expected not more than that. I do not think that 5 minutes is a long time to get feedback on PR :-) 

(2) I was not aware of the minio overlay folder ... this can be used although I see that folder as overkill for this test. All of that deployment is happening in a single line in the test in https://github.com/ViaQ/loki-operator/pull/70/files#diff-a9490ebdf9deb43030add61aafbe50aaa1b44c3db5569fcf3dfece82a95bf09aR45 and there is a lot of code that handles initial bucket deployment and minio-client that is missing from that ` overlay folder`

(4) Yes, totally agree ... this was one of the areas I thought need to be handled quickly once this merged so https://github.com/ViaQ/loki-operator/pull/70/files#diff-a9490ebdf9deb43030add61aafbe50aaa1b44c3db5569fcf3dfece82a95bf09aR36 need to move to use the gateway ... sure. Actually, this highlights the value of such a test.

(3) I am not against golang, but this is not efficient for writing such test codes ... we can discuss the conversion of this into golang if you think that this is correct ... just to explain the development of this as a framework in golang will be a significant effort and I am not sure that this is goof use of time.

eranra:
@periklis please advice **here** on next steps for this PR

periklis:
> @periklis please advice **here** on next steps for this PR

We will postpone this until we are done with the work in the lokistack-gateway as per having more focus then to address issues like providing a golang-based e2e foundation.

eranra:
@periklis ... ok ... make sense ... when you have a plan for this ... ping and we will enhance / improve etc.

PR #99, opened 2021-10-26
LOG-1840: Dev Preview Docs on how to get started with hacking on loki-operator

sasagarw:
This PR:
- Adds doc for hacking on loki operator using kind.
- Adds doc for hacking on loki operator on openshift.
- Adds some basic troubleshooting for the above two.

/cc @periklis 

### Links
- Jira: https://issues.redhat.com/browse/LOG-1840

### DIFF
--- a/docs/hack_loki_operator.md
+++ b/docs/hack_loki_operator.md
@@ -29,6 +29,8 @@ Loki Operator is the Kubernetes Operator for [Loki](https://grafana.com/docs/lok
   ```console
   kubectl get pods
   ```
+  
+  You should see `controller-manager-xxxx` and `minio-xxxx` pods running.
 
 * Now create a LokiStack instance to get the various components of Loki up and running:
 
@@ -38,7 +40,7 @@ Loki Operator is the Kubernetes Operator for [Loki](https://grafana.com/docs/lok
 
   This will create `distributor`, `compactor`, `ingester`, `querier`, `query-frontend` and `lokistack-gateway` components.
 
-  Confirm that all are up and running for `deployments` using:
+  Confirm that all components are up and running for `deployments` using:
 
   ```console  
   kubectl rollout status deployment/<DEPLOYMENT_NAME>
@@ -62,9 +64,7 @@ Loki Operator is the Kubernetes Operator for [Loki](https://grafana.com/docs/lok
   kubectl get statefulsets
   ```
 
-  _Note:_  `lokistack-gateway` is an optional component deployed as part of Loki Operator. It provides secure access to Loki's distributor (i.e. for pushing logs) and query-frontend (i.e. for querying logs) via consulting an OAuth/OIDC endpoint for the request subject.
-
-  If you don't want `lokistack-gateway` component then you can skip it by removing the `--with-lokistack-gateway` args from the `controller-manager` deployment:
+  If you don't want `lokistack-gateway` component [1] then you can skip it by removing the `--with-lokistack-gateway` args from the `controller-manager` deployment:
 
   ```console
   kubectl edit deployment/controller-manager
@@ -86,7 +86,11 @@ To cleanup deployments of the operator, you can use:
 make undeploy
 ```
 
-It will undeploy controller from the configured Kubernetes cluster in ~/.kube/config
+It will undeploy controller from the configured Kubernetes cluster in [~/.kube/config](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable)
+
+### Notes
+
+[1] `lokistack-gateway` is an optional component deployed as part of Loki Operator. It provides secure access to Loki's distributor (i.e. for pushing logs) and query-frontend (i.e. for querying logs) via consulting an OAuth/OIDC endpoint for the request subject.
 
 ## Hacking on Loki Operator on OpenShift
 
@@ -99,7 +103,7 @@ It will undeploy controller from the configured Kubernetes cluster in ~/.kube/co
 
 ### Installation of Loki Operator
 
-* Build and push the container image and then deploy the operator with:
+* Build and push the container image [2] and then deploy the operator with:
 
   ```console
   make olm-deploy REGISTRY_ORG=$YOUR_QUAY_ORG VERSION=$VERSION
@@ -132,7 +136,7 @@ It will undeploy controller from the configured Kubernetes cluster in ~/.kube/co
 * Once the object storage secret is created, you can now create a LokiStack instance to get the various components of Loki up and running:
 
   ```console
-  kubectl apply -f hack/lokistack_gateway_dev.yaml
+  kubectl -n openshift-logging apply -f hack/lokistack_gateway_dev.yaml
   ```
 
   This will create `distributor`, `compactor`, `ingester`, `querier`, `query-frontend` and `lokistack-gateway` components.
@@ -140,39 +144,37 @@ It will undeploy controller from the configured Kubernetes cluster in ~/.kube/co
   Confirm that all are up and running for `deployments` using:
 
   ```console  
-  kubectl rollout status deployment/<DEPLOYMENT_NAME>
+  kubectl -n openshift-logging rollout status deployment/<DEPLOYMENT_NAME>
   ```  
 
   where `<DEPLOYMENT_NAME>` is the name of the deployment and can be found using:
 
   ```console
-  kubectl get deployments
+  kubectl -n openshift-logging get deployments
   ```
 
   Confirm that all are up and running for `statefulsets` using:
 
   ```console  
-  kubectl rollout status statefulset/<STATEFULSET_NAME>  
+  kubectl -n openshift-logging rollout status statefulset/<STATEFULSET_NAME>  
   ```  
 
   where `<STATEFULSET_NAME>` is the name of the statefulset and can be found using:
 
   ```console
-  kubectl get statefulsets
+  kubectl -n openshift-logging get statefulsets
   ```
 
-  _Note:_  `lokistack-gateway` is an optional component deployed as part of Loki Operator. It provides secure access to Loki's distributor (i.e. for pushing logs) and query-frontend (i.e. for querying logs) via consulting an OAuth/OIDC endpoint for the request subject.
-
-  If you don't want `lokistack-gateway` component then you can skip it by removing the `--with-lokistack-gateway` args from the `loki-operator-controller-manager` deployment:
+  If you don't want `lokistack-gateway` component [1] then you can skip it by removing the `--with-lokistack-gateway` args from the `loki-operator-controller-manager` deployment:
 
   ```console
-  kubectl edit deployment/loki-operator-controller-manager
+  kubectl -n openshift-logging edit deployment/loki-operator-controller-manager
   ```
 
   Delete the flag `--with-lokistack-gateway` from the `args` section and save the file. This will update the deployment and now you can create LokiStack instance using:
 
   ```console
-  kubectl apply -f hack/lokistack_dev.yaml
+  kubectl -n openshift-logging apply -f hack/lokistack_dev.yaml
   ```
 
   This will create `distributor`, `compactor`, `ingester`, `querier` and `query-frontend` components only.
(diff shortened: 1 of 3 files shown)


docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+## Hacking on Loki Operator on OpenShift
+
+### Requirements
+
+* Install [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) or [Openshift CLI](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html) for communicating with the cluster. The guide below will be using `kubectl` for the same.
+* Create a running OpenShift cluster on AWS.
+* A container registry that you and your OpenShift cluster can reach. We recommend  [quay.io](https://quay.io/signin/).
+* Create an [S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-overview.html) in one of the AWS Regions.
+
+### Installation of Loki Operator
```
eranra, line 100: A lot of those instructions are a replication of the instructions above. For maintenance simplicity of this document please consolidate the instructions so they appear only once, and you just refer to the relevant instructions in the duplicated location.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+
+* Now create a LokiStack instance to get the various components of Loki up and running:
+
+  ```console
+  kubectl apply -f hack/lokistack_gateway_dev.yaml
+  ```
+
+  This will create `distributor`, `compactor`, `ingester`, `querier`, `query-frontend` and `lokistack-gateway` components.
+
+  Confirm that all are up and running for `deployments` using:
```
eranra, line 41: s/all/all components

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+  If you don't want `lokistack-gateway` component then you can skip it by removing the `--with-lokistack-gateway` args from the `controller-manager` deployment:
+
+  ```console
+  kubectl edit deployment/controller-manager
+  ```
+
+  Delete the flag `--with-lokistack-gateway` from the `args` section and save the file. This will update the deployment and now you can create LokiStack instance using:
+
+  ```console
+  kubectl apply -f hack/lokistack_dev.yaml
```
eranra, line 76: I see in lokistack_dev.yaml reference to https://github.com/ViaQ/loki-operator/blob/master/hack/lokistack_dev.yaml#L11 **gp2** storage class. This doesn't work with kind

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Requirements
+
+* Install [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) or [Openshift CLI](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html) for communicating with the cluster. The guide below will be using `kubectl` for the same.
+* Create a running Kubernetes cluster using kind.
+* A container registry that you and your Kubernetes cluster can reach. We recommend  [quay.io](https://quay.io/signin/).
+
+### Installation of Loki Operator
+
+* Build and push the container image and then deploy the operator with:
```
eranra, line 17: @sasagarw to me, hacking includes how to run the Loki operator from my laptop "local process" and not push the images to a remote location and load into Kind or OCP. 

What I do in order to develop is similar to shat you explain here but allows me a much quicker process because I do not need to push and pull images all the time ... also, this allows me to use a debugger to fix problems and progress with the code. Can you explain how to do that with Kind ... this is much more realistic than using `make deploy`. I have to say that the final stage after I am happy with the code is exactly what you explain here, I push and test ... but this is something that I typically do just very close to code ready.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+  ```
+
+  where `$YOUR_QUAY_ORG` is your personal [quay.io](http://quay.io/) account where you can push container images.
+
+  The above command will deploy the operator to your active Kubernetes cluster defined by your local [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/). The operator will be running in the `default` namespace.
+
+* You can confirm that the operator is up and running using:
+
+  ```console
+  kubectl get pods
```
eranra, line 30: Can you add what pods I should expect to see?

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+  kubectl rollout status statefulset/<STATEFULSET_NAME>  
+  ```  
+
+  where `<STATEFULSET_NAME>` is the name of the statefulset and can be found using:
+
+  ```console
+  kubectl get statefulsets
+  ```
+
+  _Note:_  `lokistack-gateway` is an optional component deployed as part of Loki Operator. It provides secure access to Loki's distributor (i.e. for pushing logs) and query-frontend (i.e. for querying logs) via consulting an OAuth/OIDC endpoint for the request subject.
```
eranra, line 65: I suggest moving this note to the end of the doc

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,165 @@
...
+* You can confirm that the operator is up and running using:
+
+  ```console
+  kubectl -n openshift-logging get pods
+  ```
+
+* Now you need to create a storage secret for the operator. This can be done using:
+
+  ```
+  make olm-deploy-example-storage-secret
```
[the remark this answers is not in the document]
Red-GV: @sasagarw Here is an example: https://github.com/observatorium/loki-benchmarks/blob/master/hack/create-s3-bucket.sh

docs/hack_loki_operator.md
Red-GV: Are you referring to something along the lines of a `make run` option? If so, wouldn't it just be better to have a PR which adds the option and then add the command to the docs here versus adding all the instruction here?

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+
+  ```console
+  ./hack/deploy-example-secret.sh openshift-logging
+  ```
+
+  This secret will be available in openshift-logging namespace. You can check the `hack/deploy-example-secret.sh` file to check the content of the secret.
+
+* Once the object storage secret is created, you can now create a LokiStack instance to get the various components of Loki up and running:
+
+  ```console
```
ronensc, line 134: I went through these instructions and had to add `-n openshift-logging` to make them work.

ronensc (commented):
This guide is very useful to me. thanks!

docs/hack_loki_operator.md
periklis: @eranra For now we will support only `kind` and `openshift`. The full cycle with using images is that you ensure that what you build runs also in containers. A personal registry is maybe the tax to pay, but nowadays everybody has simple access to one.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Cleanup
+
+To cleanup deployments of the operator, you can use:
+
+```console
+make undeploy
+```
+
+It will undeploy controller from the configured Kubernetes cluster in ~/.kube/config
```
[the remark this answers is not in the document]
periklis: @ronensc The answers lies in between. `kubectl` considers both locations. First on `$KUBECONFIG` and then into `~/.kube/config`. For simplicity we can reference from here to official docs and not fully replicate them here.

docs/hack_loki_operator.md
periklis: Good catch!

docs/hack_loki_operator.md
eranra: Hi @periklis I do not have any problem with supporting only kind and OCP. Totally agree with that and actually, this should be good enough for most cases, so this is good ... my point is that the development process for operators requires changes to the code and checking that it works ... iteratively, and for that to be efficient, running the operator as a local process is something we are doing a lot. (i.e. `make run` ). I suggest extending the hacking explanation to talk on `make run` and how to use that process. To me, this is the most common way to "hack" the Loki operator, and very important to explain how to do that.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+* You can confirm that the operator is up and running using:
+
+  ```console
+  kubectl get pods
+  ```
+
+* Now create a LokiStack instance to get the various components of Loki up and running:
+
+  ```console
+  kubectl apply -f hack/lokistack_gateway_dev.yaml
```
jpinsonneau, line 36: that one uses `storageClassName: gp2` that is not available with KIND

needed to use lokistack_dev.yaml and manually delete pvcs to make it works

docs/hack_loki_operator.md
Red-GV: Think this is now resolved due to the changes brought in by #101.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+It will undeploy controller from the configured Kubernetes cluster in ~/.kube/config
+
+## Hacking on Loki Operator on OpenShift
+
+### Requirements
+
+* Install [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) or [Openshift CLI](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html) for communicating with the cluster. The guide below will be using `kubectl` for the same.
+* Create a running OpenShift cluster on AWS.
```
NissesSenap, line 96: You don't have to create OCP on AWS to run this operator. Since this is a development document I would point towards CRC.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+* A container registry that you and your OpenShift cluster can reach. We recommend  [quay.io](https://quay.io/signin/).
+* Create an [S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-overview.html) in one of the AWS Regions.
```
NissesSenap, line 98: I would point to minio `kubectl apply -k config/overlays/development/minio`

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,222 @@
...
+## Hacking on Loki Operator using kind
+
+[kind](https://kind.sigs.k8s.io/docs/user/quick-start/) is a tool for running local Kubernetes clusters using Docker container "nodes". kind was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
+
+### Requirements
+
+* Install [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) or [Openshift CLI](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html) for communicating with the cluster. The guide below will be using `kubectl` for the same.
+* Create a running Kubernetes cluster using kind.
+* A container registry that you and your Kubernetes cluster can reach. We recommend  [quay.io](https://quay.io/signin/).
+
```
NissesSenap, line 14: To use the operator you currently have to have a S3 compatible storage solution.
For kind i would suggest using: `kubeclt apply -k config/overlays/development/minio/`

NissesSenap (commented):
As mentioned by other people I agree to use `make run` to make this document easier.

There is no mentions about S3 storage under kind Requirements and there should probably be suggested usage of minio.

There is no mentions of `hack/addons.yaml` which is needed when verifying this operator to actually see some data in many cases.

docs/hack_loki_operator.md
sasagarw: Although both are doing few common steps but since we refer to both (kind and openshift) as separate topics so this looks much cleaner and step-by-step process. Consolidating them and then refering to duplicate section would make bad experience IMO.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,280 @@
+# Hacking on Loki Operator
+
+Loki Operator is the Kubernetes Operator for [Loki](https://grafana.com/docs/loki/latest/) provided by the Red Hat OpenShift engineering team.
```
periklis, line 3: Please remove the reference `provided by the Red Hat OpenShift engineering team` here.

docs/hack_loki_operator.md
periklis: @sasagarw This should be `lokistack_dev.yaml` and not `lokistack_gateway_dev.yaml`. The gateway cannot be used on kind because if lacks a default ingress controller.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,280 @@
...
+  You should see `controller-manager-xxxx` and `minio-xxxx` pods running.
+
+* Now you need to create a gateway secret [2] for the operator. This can be done using:
+
+  ```console
+  kubectl create secret generic test1 \
+    --from-literal=clientID="<CLIENT_ID>" \
+    --from-literal=clientSecret="<CLIENT_SECRET>" \
+    --from-literal=issuerCAPath="<ISSUER_CA_PATH>"
+  ```
```
periklis, line 42: We can drop this as per now on kind we should use `lokistack_dev.yaml` that is not providing the gateway.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,280 @@
...
+
+* Now create a LokiStack instance to get the various components of Loki up and running:
+
+  ```console
+  kubectl apply -f hack/lokistack_gateway_dev.yaml
+  ```
+
+  This will create `distributor`, `compactor`, `ingester`, `querier`, `query-frontend` and `lokistack-gateway` components.
```
periklis, line 50: Remove reference to `lokistack-gateway` on kind for now.

docs/hack_loki_operator.md
periklis: TBH, we don't have much experience with CRC and its single-node limitations. We will keep this as AWS for now.

docs/hack_loki_operator.md
periklis: We would like to keep minio out of OpenShift deployments, because we might later address native OpenShift storage support.

docs/hack_loki_operator.md
```diff
@@ -0,0 +1,247 @@
...
+  ```console
+  kubectl get pods
+  ```
+  
+  You should see `controller-manager-xxxx` and `minio-xxxx` pods running.
+
+* Now create a LokiStack instance to get the various components of Loki up and running:
+
+  ```console
+  kubectl apply -f hack/lokistack__dev.yaml
```
periklis, line 38: Double underscore in the filename here.

sasagarw: Such a silly mistake. Updated it.

docs/hack_loki_operator.md
NissesSenap: I understand this but my understanding of this document is that it should be targeted towards developers especially since you are targeting kind in the other example in this document.
Sadly it's more or less impossible to run OCS on a single node cluster due to that CEPH loves resources.

So I really hope that you would rethink this decision and write instructions for how to deploy the operator locally without being forced to run AWS. Else there will be only people with full size clusters so they can deploy OCS or easy access to AWS that can develop this operator.

PR #100, opened 2021-10-27
LOG-1874: lokistack-gateway component restarting too many times on OpenShift

sasagarw:
This PR:
- fixes serviceAccount issue of gateway for static, dynamic modes. Currently, we were setting ServiceAccountName for all the 3 modes but this is not correct as we don't create any ServiceAccount for static and dynamic modes. Hence the deployment errors out saying missing ServiceAccountName and a reconcile error occurs. Hence moved this to openshift specific deployment configuration (this was done in #101)
- degraded condition on the lokistack was getting set but the deployments were still taking places because of the non-nil error returned by `SetDegradedCondition` function.
- fixed the `uuid` and `cookieSecret` issue by fetching the existing tenants.yaml configuration and using the same value in subsequent reconcile. Both uuid and cookieSecret will be created only once.

/cc @periklis 

### Links
- Jira: https://issues.redhat.com/browse/LOG-1874

### DIFF
--- a/internal/handlers/internal/gateway/tenant_configmap.go
+++ b/internal/handlers/internal/gateway/tenant_configmap.go
@@ -1,0 +1,90 @@
+package gateway
+
+import (
+	"context"
+
+	"github.com/ViaQ/loki-operator/internal/manifests/openshift"
+
+	"github.com/ViaQ/logerr/log"
+
+	"github.com/ViaQ/logerr/kverrors"
+	"github.com/ViaQ/loki-operator/internal/manifests"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/json"
+	"sigs.k8s.io/yaml"
+
+	"github.com/ViaQ/loki-operator/internal/external/k8s"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// LokiGatewayTenantFileName is the name of the tenant config file in the configmap
+	LokiGatewayTenantFileName = "tenants.yaml"
+)
+
+type tenantsConfigJSON struct {
+	Tenants []tenantsSpec `json:"tenants,omitempty"`
+}
+
+type tenantsSpec struct {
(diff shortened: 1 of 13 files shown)


sasagarw:
@periklis can we do same with the `cookieSecret` as well?

bundle/manifests/loki-operator.clusterserviceversion.yaml
```diff
@@ -45,7 +45,7 @@ metadata:
     operatorframework.io/arch.amd64: supported
     operatorframework.io/arch.ppc64le: supported
     operatorframework.io/arch.s390x: supported
-  name: loki-operator.v0.0.1
+  name: loki-operator.v0.1.98
```
periklis, line 48: Can we drop this random version please?

bundle/manifests/loki-operator.clusterserviceversion.yaml
```diff
@@ -594,8 +594,8 @@ spec:
                   value: quay.io/observatorium/api:latest
                 - name: RELATED_IMAGE_OPA
                   value: quay.io/observatorium/opa-openshift:latest
-                image: quay.io/openshift-logging/loki-operator:v0.0.1
-                imagePullPolicy: IfNotPresent
+                image: quay.io/sasagarw/loki-operator:v0.1.98
+                imagePullPolicy: Always
```
periklis, line 598: Same here please keep the old defaults

internal/handlers/internal/gateway/tenant_configmap.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil
+	}
+
+	tcm, err := extractTenantConfigMap(&tenantConfigMap)
+	if err != nil {
+		log.Error(err, "error occurred in extracting tenants.yaml configMap.")
+		return nil
+	}
+
+	tcmMap := make(map[string]map[string]string)
```
periklis, line 57: Suggest to make the inside map a small struct, e.g. `tenantData`

periklis (changes requested):
Looks good, only a couple of small improvements required.

periklis (approved):

PR #102, opened 2021-11-09
Update addons with ns openshift-logging

NissesSenap:
This to make it match the rest of the operator usage of openshift-logging ns

### DIFF
--- a/hack/addons.yaml
+++ b/hack/addons.yaml
@@ -5,7 +5,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: logcli
-  namespace: loki
+  namespace: openshift-logging
   labels:
     app.kubernetes.io/name: logcli
 spec:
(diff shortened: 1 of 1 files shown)


periklis:
@NissesSenap Actually this addons.yaml is out-of-date and subject for removal. The operator is dedicated to Loki as of now. This file was added at the very early stage and we haven't used it since then.

NissesSenap:
Okay, I fully understand that you only want to focus on loki. But from a development point of view isn't it good to give your self a standard way of getting data in to loki.
It will also be helpful for getting started to users. I know that you have logick for this in OCP with log-forwarder but for a kubenretes user that is not a reasonable option in many cases. 

I don't think that you should support this addons.yaml but in docs rather point it at it and say this is a easy way of sending data to your loki instance. Use some other solution in production.

If not this way do you have any other way of getting data in to loki from a development point of view @periklis?

PR #106, opened 2021-11-24
Log 1894: Updating Loki Operand to 2.4.1

Red-GV:
This PR does the following:

- Updates to Loki Operand to the 2.4.1 image
- Enables the WAL capabilities and disables the handover mechanism
- Adds the Index Gateway component
- Converts Queriers from a StatefulSet to a Deployment

The ingesters must remain as a statefulset in order to use WAL. 

JIRA: https://issues.redhat.com/browse/LOG-1894


internal/manifests/internal/config/options.go
```diff
@@ -44,9 +47,22 @@ type Parallelism struct {
...
 func (p Parallelism) Value() int32 {
 	return int32(math.Floor(float64(p.QuerierCPULimits) / float64(p.QueryFrontendReplicas)))
 }
+
+// ReplayMemoryCeiling calculates 75% of the ingester memory
+// for the ingester to use for the write-ahead-log capbability.
+func (w WriteAheadLog) ReplayMemoryCeiling() string {
+	value := int64(math.Ceil(float64(w.IngesterMemoryRequest) * float64(0.75)))
+	return fmt.Sprintf("%d", value)
+}
```
periklis, line 68: According to the official Loki production configuration (at least this is the public version of what Grafana Cloud is using), they recomend 50% (https://github.com/grafana/loki/blob/main/production/ksonnet/loki/wal.libsonnet#L17). Could you elaborate on why 75%?

internal/manifests/internal/sizes.go
```diff
@@ -148,6 +167,16 @@ var ResourceRequirementsTable = map[lokiv1beta1.LokiStackSizeType]ComponentResou
 				corev1.ResourceMemory: resource.MustParse("1Gi"),
 			},
 		},
+		WALStorage: ResourceRequirements{
+			PVCSize: resource.MustParse("150Gi"),
+		},
+		IndexGateway: ResourceRequirements{
+			PVCSize: resource.MustParse("50Gi"),
+			Requests: map[corev1.ResourceName]resource.Quantity{
+				corev1.ResourceCPU:    resource.MustParse("0.5"),
```
periklis, line 176: nit. Use the more readable form for for sub-core values: `500m`

internal/manifests/internal/sizes.go
```diff
@@ -195,6 +224,9 @@ var StackSizeTable = map[lokiv1beta1.LokiStackSizeType]lokiv1beta1.LokiStackSpec
 			Gateway: &lokiv1beta1.LokiComponentSpec{
 				Replicas: 2,
 			},
+			IndexGateway: &lokiv1beta1.LokiComponentSpec{
+				Replicas: 1,
+			},
```
periklis, line 229: Can you elaborate if we need to pin the number of replicas to 1 for the Index Gateway like we do for Compactor when we apply the default options?

internal/manifests/querier.go
```diff
@@ -46,6 +45,12 @@ func NewQuerierStatefulSet(opts Options) *appsv1.StatefulSet {
 					},
 				},
 			},
+			{
+				Name: storageVolumeName,
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
+			},
```
periklis, line 53: This is not required if we use the index-gateway as the queriers don't need the data directory anymore, same applies for the volumeMount.

internal/manifests/service_monitor.go
```diff
@@ -97,6 +98,25 @@ func NewGatewayServiceMonitor(opts Options) *monitoringv1.ServiceMonitor {
...
+
+	sm := newServiceMonitor(opts.Namespace, serviceMonitorName, l, gwEndpoint)
+
+	if opts.Stack.Tenants != nil {
+		if err := configureServiceMonitorForMode(sm, opts.Stack.Tenants.Mode, opts.Flags); err != nil {
+			return sm
+		}
+	}
+
+	return sm
```
periklis, line 117: This is replicating the lokistack gateway form of configuring the service monitor and thus not correct. The lokistack gateway has a special requirement as a multi-container pod. Apply here the pattern as for the querier in `NewQuerierServiceMonitor`.

internal/manifests/var.go
```diff
@@ -129,6 +131,11 @@ func GatewayName(stackName string) string {
 	return fmt.Sprintf("lokistack-gateway-%s", stackName)
 }
 
+// IndexGatewayName is the name of the lokiStack-index-gateway statefulset
+func IndexGatewayName(stackName string) string {
+	return fmt.Sprintf("lokistack-index-gateway-%s", stackName)
```
periklis, line 136: Use `loki` instead of `lokistack` as a prefix for now here.

periklis (commented):
Awesome work! :rocket: A couple of small improvements and we are set to go:

1. Please vet that we using explicitly the new defaults if we don't override them in Loki config or limits config (https://github.com/grafana/loki/blob/main/docs/sources/upgrading/_index.md#change-of-some-default-limits-to-common-values)
2. Same for configuration defaults (https://github.com/grafana/loki/blob/main/docs/sources/upgrading/_index.md#change-of-configuration-defaults)
3. Although we use the FIFO cache since the early beginning please vet that the configuration makes explicit usage of its configuration params (https://github.com/grafana/loki/blob/main/docs/sources/upgrading/_index.md#in-memory-fifo-caches-enabled-by-default)
4. Same for the ingester lifecycle `final_sleep` param (https://github.com/grafana/loki/blob/main/docs/sources/upgrading/_index.md#ingester-lifecycler-final_sleep-now-defaults-to-0s)
5. Same for memberlist https://github.com/grafana/loki/blob/main/docs/sources/upgrading/_index.md#memberlist-config-now-automatically-applies-to-all-non-configured-rings
6. Same for GRPC server settings https://github.com/grafana/loki/blob/main/docs/sources/upgrading/_index.md#changed-defaults-for-some-grpc-server-settings

Making the above explicit in the loki configuration enables faster lookup while troubleshooting.

internal/manifests/internal/config/options.go
Red-GV: Hmm, there appears to be a discrepancy here. The documentation recommended 75% (https://grafana.com/docs/loki/latest/operations/storage/wal/).

internal/manifests/internal/sizes.go
Red-GV: There isn't a need to keep the Index Gateway to one replica like we do for the Compactor.

api/v1beta1/lokistack_types.go
```diff
@@ -308,12 +308,19 @@ type LokiTemplateSpec struct {
...
 
-	// Gateway defines the lokistack-gateway component spec.
+	// Gateway defines the lokistack gateway component spec.
 	//
 	// +optional
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Gateway pods"
 	Gateway *LokiComponentSpec `json:"gateway,omitempty"`
+
+	// IndexGateway defines the lokistack index gateway component spec.
```
sasagarw, line 318: Nit
```suggestion
	// IndexGateway defines the index gateway component spec.
```

api/v1beta1/lokistack_types.go
```diff
@@ -416,12 +423,28 @@ type IngestionLimitSpec struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Per Stream Rate Limit"
+	PerStreamRateLimit int32 `json:"perStreamRateLimit,omitempty"`
+
+	// PerStreamRateLimitBurst defines the maximum burst bytes per stream.
+	// This value sets the upper limit for the stream rate limit. When exceeding
+	// this, the stream is limited.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="PerStreamRateLimitBurst"
```
sasagarw, line 446: Nit
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Per Stream Rate Limit Burst"
```

internal/manifests/config.go
```diff
@@ -71,6 +75,10 @@ func ConfigOptions(opt Options) config.Options {
 			QuerierCPULimits:      opt.ResourceRequirements.Querier.Requests.Cpu().Value(),
 			QueryFrontendReplicas: opt.Stack.Template.QueryFrontend.Replicas,
 		},
+		WriteAheadLog: config.WriteAheadLog{
+			Directory:             strings.TrimRight(walDirectory, "/"),
```
sasagarw, line 79: What's the significance of using `strings.TrimRight()` here?
The constant `walDirectory` already equals `/tmp/wal`. There is no `/` at the end to trim.

internal/manifests/var.go
```diff
@@ -124,6 +126,11 @@ func QueryFrontendName(stackName string) string {
 	return fmt.Sprintf("loki-query-frontend-%s", stackName)
 }
 
+// IndexGatewayName is the name of the lokiStack-index-gateway statefulset
```
sasagarw, line 129: Nit
```suggestion
// IndexGatewayName is the name of the index-gateway statefulset
```

sasagarw (changes requested):

internal/manifests/config.go
Red-GV: That was the result of copy & paste.

api/v1beta1/lokistack_types.go
```diff
@@ -416,12 +423,28 @@ type IngestionLimitSpec struct {
...
+	PerStreamRateLimit int32 `json:"perStreamRateLimit,omitempty"`
+
+	// PerStreamRateLimitBurst defines the maximum burst bytes per stream.
+	// This value sets the upper limit for the stream rate limit. When exceeding
+	// this, the stream is limited.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors="urn:alm:descriptor:com.tectonic.ui:number",displayName="Per Stream Rate Limit Burst"
+	PerStreamRateLimitBurst int32 `json:"perStreamRateLimitBurst,omitempty"`
```
periklis, line 447: Why do we introduce these two limits in this PR? Can you elaborate why this is mandatory to do now? We should change the API after having a proper discussion i.e. an enhancement proposal. Adding new knobs widens our opinionated API and we should have a good reason to do so.

internal/manifests/internal/sizes.go
periklis: Then why not having 2 replicas for HA in `1x.small` and `1x.medium`?

api/v1beta1/lokistack_types.go
Red-GV: Gotcha, sorry about that. Jumped a little ahead. These options were not in 2.2.1, but offer more opportunities to tune the ingestion rate of the cluster. So maybe it is worth the enhancement proposal. I will revert this change and just set their default values in the config file.

periklis (commented):
LGTM

sasagarw (approved):

